PR #31, opened 2023-01-30
Add :open attachments command

### LINKED ISSUE #27: Open attachments
Either a parameter to `:download`, e.g. `:download open`, or a new command `:open` that downloads and opens. "Opening" could default to `xdg-open <downloaded-file-path>` or even better have a config setting.
@ulyssa wdyt?

(I am experimenting with a new command).

benjajaja:
Fixes #27

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -10,6 +10,7 @@ use modalkit::{
 };
 
 use crate::base::{
+    DownloadFlags,
     IambAction,
     IambId,
     MessageAction,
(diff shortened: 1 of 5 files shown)


src/commands.rs
```diff
@@ -317,7 +318,26 @@ fn iamb_download(desc: CommandDescription, ctx: &mut ProgContext) -> ProgResult
...
+}
+
+fn iamb_open(desc: CommandDescription, ctx: &mut ProgContext) -> ProgResult {
+    let mut args = desc.arg.strings()?;
+
+    if args.len() > 1 {
+        return Result::Err(CommandError::InvalidArgument);
+    }
+
+    let flags = DownloadFlags::OPEN;
```
ulyssa, line 339: Can you add here:

```rust
if desc.bang {
    flags |= DownloadFlags::FORCE;
}
```

src/windows/room/chat.rs
```diff
@@ -202,29 +203,42 @@ impl ChatState {
...
                     };
 
-                    if !force && filename.exists() {
-                        let msg = format!(
-                            "The file {} already exists; use :download! to overwrite it.",
-                            filename.display()
-                        );
-                        let err = UIError::Failure(msg);
+                    if filename.exists() {
+                        if !flags.intersects(DownloadFlags::FORCE | DownloadFlags::OPEN) {
```
ulyssa, line 207: Check just for `DownloadFlags::FORCE` here, so that `:open` doesn't accidentally overwrite an already existing file.

src/windows/room/chat.rs
```diff
@@ -202,29 +203,42 @@ impl ChatState {
...
 
-                    let info = InfoMessage::from(format!(
-                        "Attachment downloaded to {}",
-                        filename.display()
-                    ));
+                    let info = if flags.contains(DownloadFlags::OPEN) {
+                        // Contrary to the docs, open::that *is* blocking (on linux with xdg-open).
+                        // https://github.com/Byron/open-rs/issues/51#issuecomment-1408261825
+                        #[allow(deprecated)]
+                        open::that_in_background(&filename);
```
ulyssa, line 231: I think you could use `open::that` here to avoid the deprecated function by doing:

```rust
let target = filename.to_os_string();
let _ = tokio::spawn_blocking(move || open::that_in_background(target));
```

tokio will then make a new blocking thread for the call, until the configured limit on the number of blocking threads is reached (512 by default).

ulyssa (changes requested):

src/commands.rs
benjajaja: my bad, fixed

src/windows/room/chat.rs
benjajaja: allright, clippy told me to explicitly `std::mem::drop` the `_` here. I am not sure, if you think it's not necessary then I'd remove it.

ulyssa:
I've merged this into main. Thank you for adding the new command!

benjajaja:
Thank you for helping me out!

PR #34, opened 2023-01-31
fix: attachment download flags + exists check

benjajaja:
Fix file never downloading (unless it has been downloaded in the past and using `!` force flag).

The logic should be:

* If file does not exist, or `!` force flag used, then download it
* Else if neither `!` or `:open` flag used, then error out

and then return downloaded-message or open-and-message.

I.e. `:open` should still open the file if it has already been downloaded. Otherwise the only way to open it is to use `!` and re-download it.

### DIFF
--- a/src/windows/room/chat.rs
+++ b/src/windows/room/chat.rs
@@ -215,17 +215,7 @@ impl ChatState {
                         },
                     };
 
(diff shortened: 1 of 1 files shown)


ulyssa:
Merged! Thanks for the fix!

PR #35, opened 2023-02-01
Append suffix to download filenames to avoid overwrites

benjajaja:
Attachment filenames collude fairly often. If the file exists then an
incrementally suffixed filename is picked. Examples:
image.jpg -> image-1.jpg
image -> image-1

If the suffixed filename is also already taken, then try with n+1 up to
a 1000 times.

The force flag skips this check and always downloads as the original
filename.

This means that `:open` will always download the file (with a new
suffix if it exists), and `:open!` will overwrite the original filename.
That is, `:open` cannot skip the download if the file exists, because
the filename of the previous download is unknown (it might have been
suffixed, and any file existing with original filename might actually 
be from another attachment).

### DIFF
--- a/src/windows/room/chat.rs
+++ b/src/windows/room/chat.rs
@@ -178,43 +178,41 @@ impl ChatState {
                         None => settings.dirs.downloads.clone(),
                     };
 
-                    let source = match &ev.content.msgtype {
-                        MessageType::Audio(c) => {
-                            if filename.is_dir() {
-                                filename.push(c.body.as_str());
-                            }
-
-                            c.source.clone()
-                        },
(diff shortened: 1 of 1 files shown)


benjajaja:
@ulyssa should this perhaps be toggled with a config option, e.g. `prefix_download_filenames`?

ulyssa:
I think that the better way to do this would be to try adding incremental numbers to the filename (before the filetype suffix) similar to how browser downloads work. So if `file.png` already exists, then try checking for `file-1.png`, `file-2.png`, etc., until one that doesn't exist is found.

ulyssa:
This is great! Thank you for implementing this @benjajaja !

PR #36, opened 2023-02-02
Support uploading image attachments from clipboard

benjajaja:
`:upload` argument is made optional. If omitted, then it will try to grab an image from the clipboard instead.

The user might not know this behaviour and the command does not indicate that the clipboard will be grabbed and uploaded without any confirmation. There could be sensitive information in it.

There should be some mechanism to confirm/ask the user, or await a "manual paste" (Ctrl-V? Shift-Ctrl-V?). I don't really know how to go on about "command steps" or something like that. @ulyssa wdyt?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -240,6 +240,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "349f9b6a179ed607305526ca489b34ad0a41aed5f7980fa90eb03160b69598fb"
 
 [[package]]
+name = "bit_field"
+version = "0.10.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dc827186963e592360843fb5ba4b973e145841266c1357f7180c43526f2e5b61"
+
+[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -588,6 +594,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-deque"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ce6fd6f855243022dcecf8702fef0c297d4338e226845fe067f6341ad9fa0cef"
+dependencies = [
+ "cfg-if",
+ "crossbeam-epoch",
+ "crossbeam-utils",
+]
+
+[[package]]
 name = "crossbeam-epoch"
 version = "0.9.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


ulyssa:
I think the way that this should probably work is through the `"*` register. When it's an image, iamb can then ask the user with a y/N prompt if they want to upload it. I just added support for using `"*` and `"+` in ulyssa/modalkit#31, but it will still need ulyssa/modalkit#83.

benjajaja:
That's really nice, the vim way.

ulyssa:
I've released a version of modalkit that includes ulyssa/modalkit#83 and just pushed #45 which includes two examples of how to use [`PromptYesNo`](https://docs.rs/modalkit/0.0.15/modalkit/input/dialog/struct.PromptYesNo.html):

- [Redacting a message](https://github.com/ulyssa/iamb/blob/a5c25f248741927b352463df7d2e9d08354edaa3/src/windows/room/chat.rs#L315-L323)
- [Leaving a room](https://github.com/ulyssa/iamb/blob/a5c25f248741927b352463df7d2e9d08354edaa3/src/windows/room/mod.rs#L223-L240)

If you add a match for `EditError::Register(RegisterError::ClipboardImage(data))` [here](https://github.com/ulyssa/iamb/blob/a5c25f248741927b352463df7d2e9d08354edaa3/src/windows/room/chat.rs#L599-L611) and turn that into a `SendAction::UploadImage(ImageData<'static>)` variant that the user is prompted to run.

benjajaja:
Thanks @ulyssa, those two things have been super easy to use!
Arboard doesn't provide the mime type, and I have tried out two mime sniffing crates to no avail, only this `image::DynamicImage::ImageRgba8` stuff works. Not even `image::guess_format` worked.
But this could easily be swapped out later if someone figures this out.

I have tried this PR with png and jpeg images.

benjajaja:
Reading the [arboard docs](https://docs.rs/arboard/latest/arboard/struct.ImageData.html) again, there is no mime-type because the bytes are always RGBA8, so the `image` crate use to convert it to png seems correct.

ulyssa:
Merged! Thank you for implementing this, @benjajaja !

schrmh:
The documentation for this is a bit unhelpful imo.
»Clipboard images can be uploaded by pasting from the clipboard register "*p, and following the confirmation dialog.«
Took me a bit to figure out that I have to write `"+p` to (usually) paste images on X11 instead of `:upload`.

E.g. this would be more helpful:
»To upload information from the clipboard, registers are used.
- Write "*p to paste a PNG image on Windows or content from the PRIMARY selection (usually contains last selected text) on X11.
- Write ":+p to paste from the CLIPBOARD selection on X11.
Follow the confirmation dialog.«

ulyssa:
@schrmh Thank you for catching that this was underdocumented! I've updated [the website](http://iamb.chat/messages/index.html#sending) to describe both registers, and provide example descriptions of behaviour.

PR #51, opened 2023-03-09
Add NetBSD install instructions

0323pin:
Hi,

I've stumbled upon this application yesterday when browsing crates.io, I've been looking for a Matrix client in Rust for quite sometime, thank you!

I've just created a package for pkgsrc and merged it into the main branch, http://mail-index.netbsd.org/pkgsrc-changes/2023/03/09/msg270698.html

If one is using pkgsrc-current it's available now, else the 2023Q1 branch is just around the corner, so in a few weeks from now, it will be available for anyone on the stable branch as well.

2023-03-09-103036_1366x768_scrot

Regards,

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,14 @@ Install Rust and Cargo, and then run:
 cargo install --locked iamb
 ```
 
+### NetBSD
+
+On NetBSD a package is available from the official repositories. To install it simply run:
+
+```
+pkgin install iamb
+```
+
 ## Configuration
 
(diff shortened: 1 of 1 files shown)


ulyssa:
@0323pin That's amazing, thank you! I've also updated the website to include install instructions.

0323pin:
Thank you for the tool :)
As said, I've been looking for this for quite sometime.

Thanks for the quick update of the site info.
I've noticed you did work at Joyent, so you should know  pkgsrc is not only for NetBSD.

Hopefully, it will build on other platforms also and will be used by others. I'm very happy with it.

PR #52, opened 2023-03-10
Create release profile with LTO

paulcarroty:


ulyssa:
Enabling LTO is a good idea. Thanks for the PR! I've removed the `debug = 1` part though. I'm okay with the build time going up in the release builds, but it adds 100MB+ to the build size.

paulcarroty:
Sure, had no idea about +100 MB size.

PR #73, opened 2023-03-25
Add Nix flake

benjajaja:
Since it's a flake, it should rely on Cargo.lock and not require maintenance, except when a new system dependency is added, which is rare with cargo. Currently the only ones are openssl and pkg-config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,12 @@ On Arch Linux a package is available in the Arch User Repositories (AUR). To ins
 paru iamb-git
 ```
 
+### Nix / NixOS (flake)
+
+```
+nix profile install "github:ulyssa/iamb"
+```
+
 ## Configuration
 
 You can create a basic configuration in `$CONFIG_DIR/iamb/config.json` that looks like:
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,94 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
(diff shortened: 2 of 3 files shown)


ulyssa:
Thanks for adding this!

Is there a reason to use `pkgs.rust-bin.nightly."2023-03-17".default` over `pkgs.rust-bin.stable.latest.default` like you had it initially? I don't know much about Nix, but I would assume that `stable` would update automatically over time, right?

benjajaja:
The reason was that .rs files are formatted with `rust-fmt` of nightly. Or at least that's what's happening with my nvim+lsp+rust installation. Also perhaps iamb used to require nightly to build some time ago? I don't remember exactly.

Anyway, yay iamb builds against stable, changed that, but only left nightly for the devShell feature. The devShell is pinned to some random recent nightly build so that it caches and doesn't recreate a new derivation every day when a new nightly comes out.

ulyssa:
Ah, that makes sense. Yeah, you'd need a nightly build for `rust-fmt`.

I've merged this into `main`. Thanks for adding this! :) I'll update the website to mention it when I do the next release.

meain:
FYI, I'm maintaining it in [nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-23.05/pkgs/applications/networking/instant-messengers/iamb/default.nix) as well.

PR #87, opened 2023-04-25
Fix error message for undefined download directory

mpldr:
The current error message is
	thread 'main' panicked at 'no dirs.download value configured!', src/config.rs:327:14
	note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

which is not only unelegant, but also wrong. The key iamb expects is `dirs.downloads`, which is now also reflected in the error message.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -324,7 +324,7 @@ impl Directories {
         let downloads = self
             .downloads
             .or_else(dirs::download_dir)
-            .expect("no dirs.download value configured!");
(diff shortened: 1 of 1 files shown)


ulyssa:
Thank you for fixing this! Out of curiosity, what platform are you on, where the `dirs::download_dir` call is failing to create a path?

mpldr:
I'm on Linux. Should I run it with backtrace?

kalloc:
xdg-user-dirs-update would helps

PR #88, opened 2023-04-25
Add manual pages

mpldr:
For reasoning, see commit messages.


ulyssa:
Thank you for writing up these man pages, @mpldr ! They look great! I'll put them in tarballs in the next release.

I replaced `scdoc` with the `mandown` crate in `build.rs` so that they should be straightforward to generate on macOS, too.

Issue #43, opened 2023-02-23
Links should be "openable"


benjajaja:
Most links are longer than the chat-message-grid-column, and thus are broken up into the next line. Then, they cannot be opened by clicking (most modern terminals have such a feature), nor can they be selected easily for copying and pasting.

If a message has no attachments and `:open` is used, then do a simple match for `^http(s)?://.+` or something like that, and pass a match to the `open` crate, it should open the browser.

ptman:
weechat has raw-mode (M-L), which is quite handy for links

benjajaja:
@ptman what does that do? Something like just printing raw chat lines?

ulyssa:
Implemented in df3148b. I made it so that `<a>` tags get a `[0]`, `[1]`, etc. shown after them, so that it's clear when someone's used markdown to link to something, and I've made it so that running `:open` will open a prompt showing the URLs and asking which number (or letter if you have enough links) you want to open.

If the message is plaintext only, then `:open` will try to parse the message body to see if it's just a URL. There might be room for improvement here, but I imagine that this should be a good step forward for now.

PR #95, opened 2023-05-04
Support configuring which program `:open` runs

benjajaja:
Allow setting "open_command" to array of [program, args...] as alternative to using `open::that`.
The `open-rs` crate now has [`open::with_command`](https://docs.rs/open/4.1.0/open/fn.with_command.html), but that only allows adding args _after_ the path arg. In my case I need `firefox --new-window path` and that seems like the most common case.

@ulyssa the two fns in chat.rs feel a bit randomly placed, but I didn't want to decide on anything about files/modules in this PR. Feel free to suggest some better organization.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -237,6 +237,7 @@ pub struct TunableValues {
     pub typing_notice_display: bool,
     pub users: UserOverrides,
     pub default_room: Option<String>,
+    pub open_command: Option<Vec<String>>,
 }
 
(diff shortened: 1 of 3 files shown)


ulyssa:
Thank you for adding this! I know there were people asking for it in `#iamb-users`.

I'm fine with the placement of the two functions for now. I might split up `chat.rs` when I get back to working on #43 again.

PR #105, opened 2023-05-18
Upload artifacts built in GitHub Actions

paulcarroty:
This is useful for testing unstable features and hotfixes.


paulcarroty:
@ulyssa still thinking to make these as independent cacheless tasks, it will not slow down the tests and will be immutable to corrupted caches, mentioned in #107.

ulyssa:
Thank you for adding this @paulcarroty ! This will be a nice way for people to get their hands on binaries to try before releases.

PR #108, opened 2023-05-20
Add support for previewing images in room scrollback

benjajaja:
An image says more than a thousand words:
image

There has been some work to get some small change into ratatui that allows for "skipping" of rendering some cells (characters). This in turn allowed to make a crate [ratatui-image](https://github.com/benjajaja/ratatui-image) that provides a image widgets for ratatui, taking care of the terminal's graphics protocol automatically. Currently the supported "protocols" are sixels, kitty, and unicode-halfblocks. iTerm2's protocol should be easy to add, I just don't have access to a mac.

This PR uses ratatui-image widget's `render()` method to render images in the `ScrollBack` widget, in the chat lines. This seems hacky, but this kind of usage was actually intended and recommended for these kind of use cases in the discussion on the ratatui PR.

Currently, the download and image processing happens in one spawned tokio task per attachment. I also have a branch where only one "worker" is created, but I don't actually know if that is an actual improvement.  I mainly made that branch/version to learn a bit more about tokio/tasks/threads in rust.

### DIFF
--- a/src/worker.rs
+++ b/src/worker.rs
@@ -244,16 +244,20 @@ async fn load_older_one(
     }
 }
 
-async fn load_insert(
-    client: &Client,
-    room_id: OwnedRoomId,
-    res: MessageFetchResult,
-    store: AsyncProgramStore,
-) {
+async fn load_insert(room_id: OwnedRoomId, res: MessageFetchResult, store: AsyncProgramStore) {
     let mut locked = store.lock().await;
-    let ChatStore { need_load, presences, rooms, .. } = &mut locked.application;
+    let ChatStore {
+        need_load,
+        presences,
+        rooms,
+        worker,
+        picker,
+        settings,
+        ..
+    } = &mut locked.application;
     let info = rooms.get_or_default(room_id.clone());
     info.fetching = false;
+    let client = &worker.client;
 
     match res {
         Ok((fetch_id, msgs)) => {
@@ -270,7 +274,14 @@ async fn load_insert(
                         info.insert_encrypted(msg);
                     },
                     AnyMessageLikeEvent::RoomMessage(msg) => {
-                        info.insert(msg);
+                        info.insert_with_preview(
+                            room_id.clone(),
+                            store.clone(),
+                            *picker,
+                            msg,
+                            settings,
+                            client.media(),
+                        );
                     },
                     AnyMessageLikeEvent::Reaction(ev) => {
                         info.insert_reaction(ev);
@@ -298,12 +309,11 @@ async fn load_older(client: &Client, store: &AsyncProgramStore) -> usize {
         .await
         .into_iter()
         .map(|(room_id, fetch_id)| {
-            let client = client.clone();
             let store = store.clone();
 
             async move {
-                let res = load_older_one(&client, room_id.as_ref(), fetch_id, limit).await;
-                load_insert(&client, room_id, res, store).await;
+                let res = load_older_one(client, room_id.as_ref(), fetch_id, limit).await;
+                load_insert(room_id, res, store).await;
             }
         })
         .collect::<FuturesUnordered<_>>()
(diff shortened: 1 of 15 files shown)


src/worker.rs
```diff
@@ -734,8 +794,22 @@ impl ClientWorker {
                     let sender = ev.sender().to_owned();
                     let _ = locked.application.presences.get_or_default(sender);
 
-                    let info = locked.application.get_room_info(room_id.to_owned());
-                    info.insert(ev.into_full_event(room_id.to_owned()));
+                    let ChatStore { rooms, previewer, .. } = &mut locked.application;
+                    let mut info = rooms.get_or_default(room_id.clone());
```
benjajaja, line 798: @ulyssa `locked.application.get_room_info(room_id.to_owned())` and `rooms.get_or_default(room_id.clone())` should behave exactly the same, right? I only changed this to comply with the borrow checker.

src/worker.rs
```diff
@@ -276,6 +296,44 @@ async fn load_older(client: &Client, store: &AsyncProgramStore) {
     }
 }
 
+fn insert_preview(
+    store: AsyncProgramStore,
+    room_id: OwnedRoomId,
+    source: PreviewSource,
+    media: Media,
+) {
```
benjajaja, line 304: Maybe there should be just one task for the previews that we send the `PreviewSource`s to with a channel.

justchokingaround:
really looking forward to this!

benjajaja:
It really only works with [this patch on Alacritty](https://github.com/microo8/alacritty-sixel), which seems to work because it simply never draw over image graphics. I'm working on fixing this in  tui-rs-revival/ratatui in the linked PR, by setting some area to be "skipped".

r3k2:
This will be great. I am using foot on wayland/sway should work. please release! :+1:

r3k2:
I was happy to see movement in this PR today, how is going? do you think if I patch this myself will it be somehow a bit stable?

benjajaja:
@r3k2 I don't know what you mean by "patch it up". I can give you a status update though:

* [x] Graphics support merged in ratatui and graphics related code has been extracted to new crate `ratatui-image`.
* [x] Iamb uses `ratatui-image` to render images with whatever protocol is detected (sixel, kitty, or unicode).
* [ ] The attachments which are "autodownloaded" for rendering are not cached in any way (this should be configurable).
* [ ] Windows builds won't work because sixel lib probably won't compile / link correctly and is missing terminal-font-size detection (rustix). We should be able to make use of the [relevant feature](https://github.com/ratatui-org/ratatui/pull/276) in ratatui for font-size, but protocol-guessing is a whole 'nother can of worms. Maybe it should guess on linux, and just have a configuration setting for windows users to e.g. force sixels.

r3k2:
@benjajaja cool, who cares about windows? :D :D just release it and say only works on Linux/Nix systems for now and that windows will come.

r3k2:
I see all check passed and such, why we are not adding this already to the main IAMB? :thinking:

ulyssa:
As @craftyguy said, this is a big feature, and I like to make sure I understand and try out PRs locally before merging, so that I can document them correctly on the website and answer questions in `#iamb-users` when they come up. 

I've been playing around with this locally and it's really cool! :tada: This is amazing stuff @benjajaja ! Thank you for doing the `halfblocks` work, too, it'll make it possible for people without sixel-aware terminals or kitty to still get image previews.

I made a couple small changes in 164503b8dc1eb5d12fe97f3c7f5dde75da52f14d: I moved the picker setup code into `src/base.rs` so that it can run as part of `ChatStore::new`, and fixed a couple issues I ran into where the `image_previews` dir didn't exist yet, and images were missing the `x` offset for placing them in the right vertical split.

I see that you just pushed the updates to use `ratatui-image@0.4.1` yesterday. If you have anything else you want to add to the PR, let me know. but otherwise I'm down to merge this, and take care of anything else in a follow-up PR.

ulyssa:
@benjajaja I've pushed some example docs to https://github.com/ulyssa/iamb.chat/pull/18 . Let me know if you think that anything else should be mentioned there.

benjajaja:
@ulyssa There is nothing else I want to add for now. At some point later `rataui-image` will get support for iTerm2's graphics protocol (I can't test it because I don't have any mac around), but that should just be version bump.

Thank you for moving the setup code!

Issue #110, opened 2023-05-25
Indicate number of members in room


ulyssa:
The room members window could list the number of members in its title.

PR #112, opened 2023-05-27
make member list title display member count

avdb13:
Fixes https://github.com/ulyssa/iamb/issues/110. Not sure how I should handle the error returned from `store.application.worker.members(room_id.to_owned())` but for now I decided to fallback to old behavior. Feel free to provide me with feedback.

### DIFF
--- a/src/windows/mod.rs
+++ b/src/windows/mod.rs
@@ -529,11 +529,16 @@ impl Window<IambInfo> for IambWindow {
 
                 Spans::from(title)
             },
-            IambWindow::MemberList(_, room_id, _) => {
+            IambWindow::MemberList(state, room_id, _) => {
                 let title = store.application.get_room_title(room_id.as_ref());
-
-                Spans(vec![bold_span("Room Members: "), title.into()])
-            },
+                let n = state.len();
+                let v = vec![
+                    bold_span("Room Members "),
+                    Span::styled(format!("({n}): "), bold_style()),
+                    title.into(),
+                ];
+                Spans(v)
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/windows/mod.rs
```diff
@@ -530,10 +530,18 @@ impl Window<IambInfo> for IambWindow {
                 Spans::from(title)
             },
             IambWindow::MemberList(_, room_id, _) => {
-                let title = store.application.get_room_title(room_id.as_ref());
-
-                Spans(vec![bold_span("Room Members: "), title.into()])
-            },
+                let v = match store.application.worker.members(room_id.to_owned()) {
```
ulyssa, line 533: Fetching the list of members is unfortunately expensive in big rooms. We don't want to do that every time we print the title, so this will need to get the length of the list from the `ListState`. I'm going to add a new method to take care of that over in ulyssa/modalkit#116, which you can then use here.

You'll also want to make sure to update the `get_win_title()` method.

ulyssa: I just pushed a commit for ulyssa/modalkit#116. If you want, you can point the `Cargo.toml` at the Git ref on Github and I'll update it once I cut a modalkit release.

avdb13:
How do you like this? I had to mess a bit around with `AsRef<str>`  to avoid having to spam `.to_string` everywhere. I like this approach.

ulyssa:
I changed the `AsRef<str>` changes back and made it write out the `Span::styled` in the method instead, since I want that function to reuse the `&str` when possible, to minimize how much copying going on every time the window redraws.

Thank you for implementing and resolving #110! :tada:

PR #117, opened 2023-06-07
Fix Nix flake build on Darwin

satoqz:
Hi there,

building iamb on darwin requires the `AppKit` and `Security` frameworks which must be explicitly specified as dependencies when using nix.

Additionally I've removed `openssl` from the `nativeBuildInputs` as it's only required in `buildInputs` and added the `result` output directory used by nix to the gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
+/result
 /TODO
 /docs/iamb.[15]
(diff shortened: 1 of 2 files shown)


benjajaja:
Just tried and linux/x86_64 works as expected, thank you for fixing that `nativeBuildInputs`!

ulyssa:
Thank you for fixing the flake build for Darwin, @satoqz , and thank you for checking that it still works on Linux, @benjajaja !

PR #144, opened 2023-08-10
chore(doc): make checkmarks to emoji to fix broken randering

wrenix:
i do not like to have :heavy_check_mark: in green and grey

renderes: https://github.com/wrenix/iamb/blob/wrenix-patch-1/README.md


benjajaja:
Does github render it differently?
One drawback of using these shortcodes is that it's worse when viewing as plain text.

wrenix:
hmm, on my c

PR #155, opened 2023-09-10
Support composing messages in an external editor

capgelka:
I found it not that difficult to implement this feature that I wanted.

If there are some issues with the current implementation, I would be glad to fix them.

PR adds a new command `:editor` that opens `$EDITOR` in the current window or in a new one (if `$EDITOR` is a GUI one),  using a new temporary file, the message from this file will be sent immediately after leaving the editor.
Also, current text from a buffer (if any) will be used as the initial content of the file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -253,9 +253,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.0.2"
+version = "2.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "487f1e0fcbe47deb8b0574e646def1c903389d95241dd1bbcc6ce4a715dfc0c1"
+checksum = "b4682ae6287fcf752ecaabbfcc7b6f9b72aa33933dc23a554d853aea8eea8635"
 
 [[package]]
 name = "blake3"
@@ -420,7 +420,7 @@ version = "4.1.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "76e21918af71fb4bcd813230cf549e33d14f73d0326b932b630ce2930332b131"
 dependencies = [
- "bitflags 2.0.2",
+ "bitflags 2.4.0",
  "clap_derive",
  "clap_lex",
  "is-terminal",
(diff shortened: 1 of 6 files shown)


Fabrice-Bernes:
> after saving the file, the message will be sent immediately.

I think it's better to send the message only **after leaving the editor**, no? Sometimes I will save what I have and keep writing.

In any case, thanks for adding this feature.

capgelka:
You are right! Hopefully, the crate I used works exactly the way you described, just didn't realize that at first, because tested it only with `:wq`. So I have fixed the description.

Fabrice-Bernes:
Hey, I compiled from your commit but `:editor` returns  `Failed command: invalid command: editor`.

capgelka:
Are you sure that you running the version from my commit? Maybe you have compiled the release but running an old debug version or vice versa.

If you are on Linux `nm target/debug/iamb | grep iamb_editor`  (or release) should succeed. If not, you are running the wrong binary.

ulyssa:
@capgelka This looks great, thank you for implementing this! I added some code to force a redraw of the screen afterwards, since at least for me locally I had leftover characters on the screen after the subprocess exited.

Fabrice-Bernes:
I hope it's ok to ask about this now that the commit has been merged,
but i'd like to know what I did wrong when trying @capgelka 's commit.
- I'm on linux, 
- ran `git clone "https://github.com/ulyssa/iamb" && cd iamb`
- then `git fetch origin pull/155/head:MASTER`
- `git checkout MASTER`
- See capgelka's changes there
- `cargo install --force --locked iamb`
- `cd ~/.cargo/bin && ./iamb`

Thanks in advance

capgelka:
I also hope that it's ok to have such a conversation, though probably better to use stack overflow for such questions anyway. 
To install a package from the current directory one has to use `.` as an argument, so you have just reinstalled `iamb` from crates.io

ulyssa:
Yep, continuing commenting here is fine!

As @capgelka said, the positional argument to `cargo install` is the name of a crate on crates.io to install. You can use `cargo install --path /path/to/iamb/repo/` to install from the directory. You can also do `cargo build --release`, and copy the binary from `./target/release/iamb`.

Fabrice-Bernes:
I should have noticed when the installation section of the README doesn't mention cloning the repo...
Anyways, i tried again, now with `--path` and it works just fine.

Thank you both for the explanation and sorry for dragging this conversation.

Issue #159, opened 2023-09-15
Support SSO authentication


captnfab:
Some servers (Eg. mozilla.org) only support [SSO auth](https://spec.matrix.org/latest/client-server-api/#single-sign-on), they cannot be accessed from iamb that expects password authentication.

ulyssa:
I've merged the work that @chlohal did in #160, which should add support for OAuth and SAML, if I understand correctly what the SDK is handling. OpenID Connect will depend on updating the SDK, and doing the work to migrate from sled to sqlite (see #61).

Issue #158, opened 2023-09-13
Support following the .well-known entries for the username's domain


ulyssa:
This has come up several times in `#iamb-users`. Some homeservers are placed under a different domain or subdomain than what's in the username, and have `.well-known` entries set up. Since `iamb` requires passing in a homeserver URL, some users have mistakenly put in the wrong value, which usually results in HTTP errors from POSTing to the login path.

I'd like to make the homeserver URL config optional, and check for the `.well-known` entry when it's not provided, fall back to building the URL from the domain, and then erroring out with a message asking the user to provide a value.

PR #160, opened 2023-09-21
Add support for logging in with SSO

chlohal:
Use `matrix-sdk`'s `sso-login` feature to provide SSO login.

Gives an interactive option at launch time for users to pick between SSO and password authentication.

In common with [the `matrix-sdk` project](https://github.com/matrix-org/matrix-rust-sdk/pull/176#issuecomment-800318936), I couldn't figure out any way to test this feature because it relies on a browser. If there's anything that I should do, please say! :)

First PR in Rust, so please feel free to criticize!!


chlohal:
Fixes #159

ulyssa:
In case anyone runs into this before I have a chance to get to #158 : if you're using `matrix.org` and SSO, the right homeserver url is `https://matrix-client.matrix.org`.

Issue #111, opened 2023-05-27
heavy cpu usage in rooms with many members


paulcarroty:
Partial workaround: using the `TOKIO_WORKER_THREADS=2` env variable.

paulcarroty:
Full workaround for Linux & cgroups:

```
sudo tee -a /etc/systemd/system/iamb.slice << 'EOF'
[Slice]
CPUQuota=10%
MemoryHigh=1G
EOF
```
Then start iamb:

```
sudo systemd-run --uid=$USER --slice=iamb.slice -t iamb
```
Got stable 10% CPU usage, no freezes/another issues.

k4r4b3y:
I see high CPU usage on my debian 11 machine, too. CPU cooling fans of my laptop go full-speed while using iamb.

ormai:
On my system iamb averages to 35% of cpu usage and 500M of ram while being inside the Welcome screen. It doesn't vary much over time.

The cpu is stable at around 70°C and the fans spin like a jet.

ulyssa:
I've merged #168 as b2b47ed. If people who've been having issues with high CPU use could verify that the change improves things for them, I'll go ahead and close this issue.  You can grab binaries from [the CI action](https://github.com/ulyssa/iamb/actions/runs/6527435745?pr=168).

PR #168, opened 2023-10-14
Reduce CPU usage by instead fetching read receipts after related sync events

Benjamin-L:
On my machine, this reduces idle CPU usage from 150% to ~0.5%.


Benjamin-L:
Fixes: #111

ulyssa:
Thank you for digging into this, @Benjamin-L !

PR #171, opened 2023-10-27
Fix not showing display names in already synced rooms

### LINKED ISSUE #149: `"username_display": "displayname"` does not work
OS: openSUSE Tumbleweed
Iamb version: 0.0.8 (installed from OBS repo [home:smolsheep](https://build.opensuse.org/project/show/home:smolsheep))
Logs (level: debug): https://paste.opensuse.org/493c0afc2a1e
Config:

```json
{
	"profiles": {
		"kvasyan": {
			"user_id": "@kvasyan:sibnsk.net",
			"url": "https://matrix.sibnsk.net"
		}
	},
	"default_profile": "kvasyan",
	"settings": {
		"log_level": "debug",
		"username_display": "displayname"
	},
	"layout": {
		"style": "restore"
	}
}
```

Thank you very much for your work! Unfortunately, i've faced one issue.

Iamb just puts Matrix ID (`@user:example.com`) instead of display names. Tried in rooms with from 10 to 500 members and even in dms.
It's very uncomfortable in rooms with bridges to other messengers, because every username transforms into sth like `@telegram_12345678910:t2bot.io`.

Thanks for attention.

benjajaja:
Fixes #149

The display_names only got populated from the
OriginalSyncRoomMemberEvent. When opening an already synced room, display_names is not populated.

ClientWorker::get_room extends the returned FetchedRoom to include RoomMembers from Room::members_no_sync. This allows RoomState::new to populate the RoomInfo's display_names.

Thanks to @Benjamin-L for finding the source of the bug in that issue. It was mentioned that it might be problematic to iterate over all members of large rooms. The largest room I had available to test this was only about 200 members, but it seemed fine.

### DIFF
--- a/src/base.rs
+++ b/src/base.rs
@@ -2,6 +2,7 @@
 //!
 //! The types defined here get used throughout iamb.
 use std::borrow::Cow;
+use std::collections::hash_map::IntoIter;
 use std::collections::{HashMap, HashSet};
 use std::convert::TryFrom;
 use std::fmt::{self, Display};
@@ -1042,6 +1043,43 @@ pub struct SyncInfo {
     pub dms: Vec<Arc<(MatrixRoom, Option<Tags>)>>,
 }
 
+bitflags::bitflags! {
+    /// Load-needs
+    #[derive(Debug, PartialEq)]
+    pub struct Need: u32 {
+        const MESSAGES = 0b00000001;
+        const MEMBERS =  0b00000010;
+    }
+}
+
+/// Things that need loading for different rooms.
+#[derive(Default)]
+pub struct RoomNeeds {
+    needs: HashMap<OwnedRoomId, Need>,
+}
+
+impl RoomNeeds {
+    /// Mark a room for needing something to be loaded.
+    pub fn insert(&mut self, room_id: OwnedRoomId, need: Need) {
+        let new_need = if let Some(mut existing_need) = self.needs.remove(&room_id) {
+            existing_need.insert(need);
+            existing_need
+        } else {
+            need
+        };
+        self.needs.insert(room_id, new_need);
+    }
+}
+
+impl IntoIterator for RoomNeeds {
+    type Item = (OwnedRoomId, Need);
+    type IntoIter = IntoIter<OwnedRoomId, Need>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.needs.into_iter()
+    }
+}
+
 /// The main application state.
 pub struct ChatStore {
     /// `:`-commands
@@ -1066,7 +1104,7 @@ pub struct ChatStore {
     pub settings: ApplicationSettings,
 
     /// Set of rooms that need more messages loaded in their scrollback.
-    pub need_load: HashSet<OwnedRoomId>,
+    pub need_load: RoomNeeds,
 
     /// [CompletionMap] of Emoji shortcodes.
     pub emojis: CompletionMap<String, &'static Emoji>,
@@ -1111,11 +1149,6 @@ impl ChatStore {
             .and_then(|i| i.name.as_ref())
             .map(String::from)
             .unwrap_or_else(|| "Untitled Matrix Room".to_string())
-    }
-
-    /// Mark a room for loading more scrollback.
-    pub fn mark_for_load(&mut self, room_id: OwnedRoomId) {
-        self.need_load.insert(room_id);
     }
 
     /// Get the [RoomInfo] for a given room identifier.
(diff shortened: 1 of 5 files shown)


Benjamin-L:
If you want a moderately large room to test with, `#rust:matrix.org` has 9,078 members. For a huge room, `#matrix:matrix.org` has 48,778.

Benjamin-L:
So I tested out this branch, and it does seem to fix the displayname bug, but it has a pretty significant perf impact. I ran a test opening the client to `#rust:matrix.org`:

2f51e968bded019ab00d302cc0f4a9c4234895e (current head of `benjajaja/fix_displayname_already_synced`)

image

8943909f06654261c948327f26519e44d7c0f0d (current head of `main`)

image

The `benjajaja/fix_displayname_already_synced` build takes noticeably longer than `main` before the UI is visible when starting up. I don't have an easy way to measure this, but subjectively it's ~2s which lines up with the period of high cpu usage. This delay also occurs when opening a large room after startup.

I haven't tested with `#matrix:matrix.org` yet because I don't want to increase my HS database size by 1000%. Assuming scale linearly with user count, and it would have a ~13s period of frozen UI and high cpu usage.

---

My only real idea for improving this would be to fetch member details on-demand for each member, rather than all at once when the room is loaded. Unfortunately, this sounds _much_ more complicated to implement, and likely prone to reintroducing the bug again in the future.

A second possibility would be to run the `members_no_sync` in the background, without blocking the UI on it. This wouldn't help with the cpu usage, but would probably help with the UI freezes.

It might be worth talking with some of the `matrix-rust-sdk` devs on to see if they have recommendations for handling rooms with large member counts. It's also probably worth noting that the bar for large-room perf is not very high. Element web handled the `#matrix:matrix.org` room very poorly last time I tried this (~1 year ago). Not sure about other clients.

ulyssa:
I think that the way to do this in order to avoid having to fetch the display names for all users in the room is to do something similar to how rooms signal that they need more messages loaded: 

- While [looping over messages](https://github.com/ulyssa/iamb/blob/8943909f06654261c948327f26519e44d7c0f0d7/src/windows/room/scrollback.rs#L1308), their senders could be placed into a `HashSet<OwnedUserId>` if we don't have their display names available
- This set then gets copied into a `HashSet<(OwnedRoomId, OwnedUserId)>` in the `ChatStore` that indicates which room/user pairs need to have their membership information fetched
- An async loop in `src/worker.rs` periodically checks this set and loads information for just those users.

This does mean that user IDs might be displayed for a second or two when scroling back and seeing messages from users whose information hasn't been fetched, but it should be much lighter on the CPU and network.

benjajaja:
Changing it to load it in worker/background.

@Benjamin-L thank you so much for perf-testing this. The current implementation is indeed unacceptably locking the UI.

@ulyssa I am not sure if we should only load displaynames for users that have messages currently in scrollback. Because then we get the "ID flashing before displayname" for any new incoming messages of a user that doesn't have some message in the recent scrollback. I would rather load `members_no_sync` at once on opening.

I also think we could use `joined_members_no_sync` which excludes invited- and left-members, unless I'm missing something.

benjajaja:
Also perhaps `room.display_name()` and `room.tags()` in worker [get_room](https://github.com/ulyssa/iamb/blob/main/src/worker.rs#L1145) could also be loaded in background, together with the members.

Benjamin-L:
> Changing it to load it in worker/background.
> 
> @Benjamin-L thank you so much for perf-testing this. The current implementation is indeed unacceptably locking the UI.

thanks :) If you push the background-loading code I should be able to run another test later today.

> @ulyssa I am not sure if we should only load displaynames for users that have messages currently in scrollback. Because then we get the "ID flashing before displayname" for any new incoming messages of a user that doesn't have some message in the recent scrollback. I would rather load members_no_sync at once on opening.

We could probably avoid that problem for incoming messages by blocking the message event handlers on a call to `get_member` if the sender is not already in the members list. Another possibility for reducing (but not eliminating) flicker would be to do something like [this](https://github.com/Benjamin-L/iamb/commit/0908bc400746cc12b906bddc630831f289224644) instead of having the worker polling on a timer.

> I also think we could use `joined_members_no_sync` which excludes invited- and left-members, unless I'm missing something.

I don't think that will work, since messages will stick around after a member leaves the room.

benjajaja:
> thanks :) If you push the background-loading code I should be able to run another test later today.

I have to do it properly, I just piggy-backed on the loading older messages thread. It does kinda work but it later re-fetches the members unnecessarily. I'll give it its own `need_load` flag or something, but I think I can still use the same task as load-older-messages.

> We could probably avoid that problem for incoming messages by blocking the message event handlers on a call to `get_member` if the sender is not already in the members list. Another possibility for reducing (but not eliminating) flicker would be to do something like [this](https://github.com/Benjamin-L/iamb/commit/0908bc400746cc12b906bddc630831f289224644) instead of having the worker polling on a timer.

I would be very hesitant to block in the message event handler. Your solution of waking-up sounds much better, thanks!

> > I also think we could use `joined_members_no_sync` which excludes invited- and left-members, unless I'm missing something.
> 
> I don't think that will work, since messages will stick around after a member leaves the room.

Ah, of course.

benjajaja:
@Benjamin-L this now has background-loading. If I understand correctly, this would play very nicely with your `notify` branch.

Benjamin-L:
Ran a test with the new code (e71e02e987b1bbadd8119f04bc0c0b869144209).
 
image

This is pretty much what we would expect: it takes a few seconds before the displaynames pop in, and we still have the cpu spike, but the UI remains responsive the whole time. After the room was loaded for the first time, I didn't see any additional cpu spikes.

ulyssa:
Thank you so much for fixing this @benjajaja , and thank you for testing its performance, @Benjamin-L !

PR #182, opened 2023-11-30
Fix truncation/padding for non-ASCII sender names

Benjamin-L:
Found out about this one when somebody with the displayname "qxofficial🍥🏳️‍⚧️" joined the rust room. This string is 30 bytes long in utf8, so iamb decided to truncate it at byte 28. This happens to be in the middle of the VS16 codepoint, which caused a panic.

`take_width` calculates the width of a grapheme cluster by adding the width of all it's code points under UAX#11. This overcounts many emoji sequences, like 🏳️‍⚧️, which consists of "🏳<ZWJ>⚧". Calculating the width of a name wrong currently causes the entire messages window UI to be misaligned, which is why this PR is marked as a draft.

The tricky bit is that the number of terminal cells that 🏳️‍⚧️ takes up varies depending on your font and terminal emulator. In some contexts, it will display as 🏳️‍⚧️ (one cell), while in others it will display as "🏳⚧" (two cells). As far as I can tell, there is no way to accurately predict the width of a given string without actually printing it to the terminal and querying the current cursor position with ANSI escapes.

The best thing I've come up with so far is:

 - pessimistically assume that the terminal/font doesn't support ZWJ sequences/country flags/skin tone modifiers/etc, and keep the current width calculation for determining truncation and _minimum_ padding.
 - emit ANSI escapes to set the cursor position after printing the displaynames, to fix the alignment issues.

### DIFF
--- a/src/message/mod.rs
+++ b/src/message/mod.rs
@@ -58,7 +58,7 @@ use crate::{
     base::{IambResult, RoomInfo},
     config::ApplicationSettings,
     message::html::{parse_matrix_html, StyleTree},
-    util::{space_span, wrapped_text},
+    util::{space, space_span, take_width, wrapped_text},
 };
 
 mod html;
@@ -898,13 +898,13 @@ impl Message {
         }
 
         let Span { content, style } = self.sender_span(info, settings);
-        let stop = content.len().min(28);
-        let s = &content[..stop];
+        let ((truncated, width), _) = take_width(content, 28);
+        let padding = 28 - width;
 
         let sender = if align_right {
-            format!("{: >width$}  ", s, width = 28)
+            space(padding) + &truncated + "  "
         } else {
-            format!("{: <width$}  ", s, width = 28)
+            truncated.into_owned() + &space(padding) + "  "
         };
 
         Span::styled(sender, style).into()
--- a/src/util.rs
+++ b/src/util.rs
@@ -26,19 +26,19 @@ pub fn split_cow(cow: Cow<'_, str>, idx: usize) -> (Cow<'_, str>, Cow<'_, str>) 
 
 pub fn take_width(s: Cow<'_, str>, width: usize) -> ((Cow<'_, str>, usize), Cow<'_, str>) {
     // Find where to split the line.
-    let mut idx = 0;
     let mut w = 0;
 
-    for (i, g) in UnicodeSegmentation::grapheme_indices(s.as_ref(), true) {
-        let gw = UnicodeWidthStr::width(g);
-        idx = i;
-
-        if w + gw > width {
-            break;
-        }
-
-        w += gw;
-    }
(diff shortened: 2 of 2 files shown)


Benjamin-L:
Yeah, the `take_width` function is "greedy" in that it always splits on the last grapheme cluster that is below the length limit. This is important for more than just avoiding a trailing "🏳". Splitting on codepoints would split "🏳\<ZWJ\>⚧" into "🏳\<ZWJ\>" and cause fun problems when you try to concatenate that with another string :)

That ratatui issue looks very relevant, but I haven't read through the whole thing yet.

benjajaja:
Aha, so it's not that simple, of course!

Here is the latest comment on that ratatui thread: 

> I just noticed a new crate (released 5 days ago): https://crates.io/crates/unicode-display-width by @jameslanska which might be a good "someone else's problem" type solution to the issue at hand.

This _might_ just solve the problem - however I hope I am not ignoring facts again :sweat_smile:

Benjamin-L:
So I finally read through https://github.com/ratatui-org/ratatui/issues/75. I believe that half of the UI alignment issues are in ratatui, but the other half is in iamb (because we're calculating padding ourselves). There's no way I'm gonna solve this problem here, and this PR is purely an improvement on the status quo so I'm unmarking this as a draft.

The results of this change are:

 - iamb no longer panics on unicode displaynames where byte 28 is not a codepoint boundary 
 - fixes a bug in `take_width` where it would truncate the string one character _after_ the maximum width and then report the incorrect width back to the caller (because the `idx = i` assignment happened before the `w + gw > width` check)

A problem that was present both before and after this change is that unicode names can result in misalignment in the UI. This can result in message text appearing in the wrong column while the rest of the UI looks okay, or it can result in the entire UI being misaligned (including ratatui window borders). Currently, the former case occurs with any string where the UTF-8 byte count does not not correspond to the display width, which is very frequent. After this change, we only get misalignment in the text column in the same situations where we get misalignment in the window borders. This is much less frequent (occurs when the display width reported by unicode-width doesn't match up to the terminal emulator's behavior), and is a _very_ thorny problem. Whatever approach works for https://github.com/ratatui-org/ratatui/issues/75 will probably also work for our padding calculations here.

ulyssa:
Oof, thank you for catching this and fixing it! Slicing the `&str` was definitely the wrong thing to do there. I've tried to be careful about using `unicode-width` and `unicode-segmentation` on user-provided text, but I'm sure there's still some places that have it wrong.

I don't think there's much that we can do when the terminal and `iamb` are using different Unicode Character Databases and disagree about what's a wide character and not, but we could possibly filter out ZWJs when printing to the terminal, and calculate widths of the non-ZWJ characters only. This might be annoying for users with terminals with ZWJ support (I think I remember the macOS Terminal handling several?), so maybe there could be a way to disable it. I'll have to read through the ratatui thread in greater depth later, and get an idea of what others are doing to solve this.

Issue #172, opened 2023-10-28
`:chats` = `:dms` + `:rooms`


foxyseta:
Element supports a unified list featuring both DMs and rooms. Perhaps that could also be useful here?

PR #184, opened 2023-12-06
Add a new `:chats` window that lists both DMs and Rooms

Vaelio:
Hello everyone,

It's my first PR in rust so feel free to criticise.

The PR adds `:chats` command, listing both DMs and Rooms. I've added a tag [DM] or [ROOM] to make sure we can differenciate if two have the same name.

image

(related to #172 )

### DIFF
--- a/src/base.rs
+++ b/src/base.rs
@@ -1172,7 +1172,7 @@ pub enum IambId {
     /// A Matrix room.
     Room(OwnedRoomId),
 
-    /// The `:rooms` window.
+    /// The `:dms` window.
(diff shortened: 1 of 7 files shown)


Vaelio:
I didn't catch the typo in the commit until now somehow, sorry for that. The program now compiles.

ulyssa:
This looks great! Thank you for taking care of #172! I've added it to the `SortValues` so that people can specifically sort this window, and moved the DM/Room label to be at the beginning of the room tags when they're present.

PR #191, opened 2023-12-31
Update README.md to add openSUSE Tumbleweed

Rerum02:


ulyssa:
Thank you for adding this to the README, @Rerum02 ! I'll make sure to put it on [iamb.chat](https://iamb.chat), too.

@photosheep that's exciting! Waiting until the next release is cut is probably a good idea; I'd like #200 to be in it, but even if I cut one before that, there's still a lot of other good features and bug fixes currently in main that will be in 0.0.9.

PR #192, opened 2024-01-03
Add support for desktop notifications

benjajaja:
Basic notifications feature.

Can be enabled in config.json, disabled by default.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -391,6 +391,12 @@ pub enum UserDisplayStyle {
     DisplayName,
 }
 
+#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq)]
+pub struct Notifications {
+    pub enabled: bool,
+    pub show_message: Option<bool>,
+}
+
 #[derive(Clone)]
 pub struct ImagePreviewValues {
     pub size: ImagePreviewSize,
@@ -476,6 +482,7 @@ pub struct TunableValues {
     pub message_user_color: bool,
     pub default_room: Option<String>,
     pub open_command: Option<Vec<String>>,
+    pub notifications: Notifications,
     pub image_preview: Option<ImagePreviewValues>,
 }
 
@@ -496,6 +503,7 @@ pub struct Tunables {
     pub message_user_color: Option<bool>,
     pub default_room: Option<String>,
     pub open_command: Option<Vec<String>>,
+    pub notifications: Option<Notifications>,
     pub image_preview: Option<ImagePreview>,
 }
 
@@ -518,6 +526,7 @@ impl Tunables {
             message_user_color: self.message_user_color.or(other.message_user_color),
             default_room: self.default_room.or(other.default_room),
             open_command: self.open_command.or(other.open_command),
+            notifications: self.notifications.or(other.notifications),
             image_preview: self.image_preview.or(other.image_preview),
         }
     }
@@ -538,6 +547,7 @@ impl Tunables {
             message_user_color: self.message_user_color.unwrap_or(false),
             default_room: self.default_room,
             open_command: self.open_command,
+            notifications: self.notifications.unwrap_or_default(),
             image_preview: self.image_preview.map(ImagePreview::values),
         }
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -214,6 +214,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5f093eed78becd229346bf859eec0aa4dd7ddde0757287b2b4107a1f09c80002"
 
 [[package]]
+name = "async-broadcast"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7c48ccdbf6ca6b121e0f586cbc0e73ae440e56c67c30fa0873b4e110d9c26d2b"
+dependencies = [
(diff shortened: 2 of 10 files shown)


paulcarroty:
Wow, nice work! Maybe worth telling about `notifications: true` in iamb config.

benjajaja:
@paulcarroty I put it in [docs/iamb.5.md](https://github.com/ulyssa/iamb/pull/192/files#diff-0464bb30be6db7f5846cb37781ee1b1f630c1e65555b500d80a571f78c80cc2c), is there somewhere else where I should add it?

benjajaja:
@ulyssa do you have some idea how I could get on about detecting if a room/dm is currently "open", to skip showing a notification in that case?

I suppose that the notification handler should only have access to `ProgramStore` and there isn't anything there that could tell if a chat is open.

BTW, it seems like `modalkit::widgets::WindowOps<_>::close` never gets called. If it would, perhaps that would be a good place to (un)track current windows, together with something else for opening.

ulyssa:
> BTW, it seems like modalkit::widgets::WindowOps<_>::close never gets called. If it would, perhaps that would be a good place to (un)track current windows, together with something else for opening.

It looks like it's missing a `WindowOps::close` call on the return value [here](https://github.com/ulyssa/modalkit/blob/2445dab464d7dcc8900b252467f1a5ad0563785e/crates/modalkit-ratatui/src/windows/layout.rs#L1377). That should get fixed, but I don't think it will be enough to detect what's open here since it only gets called when the window is actually closed and not when navigated away from with a `WindowAction::Switch`.

> do you have some idea how I could get on about detecting if a room/dm is currently "open", to skip showing a notification in that case?

There's not a great way to get this from outside the `Application` struct at the moment, since the `ScreenState`  has all of the information need to make this work right. I think there are two choices:

1. Each time `Application::redraw` is called, update a `draw_curr: Instant` field in `ChatStore`; then, when rooms are drawn, copy `draw_curr` to a `draw_last` field in the `RoomInfo`. To determine if a room is open, you then only need to check whether `draw_last == draw_curr`.
1. Delay processing messages into `Application`. There are several ways this could be done, but I'm thinking that messages could be placed in a queue in `ChatStore` and then dequeued in `Application::step` just before the `self.redraw` call, so that we always show new messages when we draw. Another way to do this would be to send fetched messages over a tokio mpsc channel and `try_recv` them in `Application`, which could have the benefit of not needing to lock the store. 

I think I like the second option the most. I've been taking a look at what handling intentional mentions could look like, and I think a similar processing mechanism could be used there. I'll try to implement this along with the current window checking, and then you can use that for notifications.

src/config.rs
```diff
@@ -476,6 +476,7 @@ pub struct TunableValues {
     pub message_user_color: bool,
     pub default_room: Option<String>,
     pub open_command: Option<Vec<String>>,
+    pub notifications: bool,
```
ulyssa, line 479: Can you make this a String instead that can be set to `"all"` or `"none"`? That way it can be configurable to other values in the future, like `"mentions"`.

ulyssa: Or rather, a string in the `config.json`, but an enum that derives `Deserialize` here.

benjajaja:
Regarding option 1 or 2, my first instinct is against processing / showing notifications in the draw-loop. The app draws in an interval now, if I'm not mistaken, and it would maybe be nice to change it so it only draws on state changes or events in the future. In that scenario, notification could just be another event to trigger a draw, but overall it sounds like these two shouldn't be coupled together.

benjajaja:
I found out that the notification settings matrix of the screenshots can be retrieved and set with the SDK, which makes a lot of sense so that the user doesn't have to manually copy it over to each client. The client basically only needs one "enable in this session" toggle. We could later add commands to set or display those settings, but for now I think it's quite okay to leave it out.

I had to add a check that the notification-timestamp is greater than the app-startup-timestamp, because I am getting a barrage of past notifications. I'm not sure how a push notification is supposed to be "marked as read", [perhaps we need to send a read-receipt](https://docs.rs/matrix-sdk/latest/matrix_sdk/ruma/api/client/push/get_notifications/v3/struct.Notification.html#structfield.read).

ulyssa:
> I found out that the notification settings matrix of the screenshots can be retrieved and set with the SDK, which makes a lot of sense so that the user doesn't have to manually copy it over to each client. The client basically only needs one "enable in this session" toggle. We could later add commands to set or display those settings, but for now I think it's quite okay to leave it out.

Yep, I agree with this. Getting notifications enabled is the first step, and eventually there can be `:notifications ...` commands for updating the shared notification settings for the room.

I noticed that the CI build failed on Windows. I've put up a fix for that in #221, and verified that it builds and notifications show up on Windows. (Although I'm not sure whether anyone is using iamb on Windows, since it looks like keyboard input got screwy after either a crossterm change or a Windows Terminal change, #220). If you pull in `main`, it should be able to pass CI.

Having tried the PR out locally, I think it's ready to merge. It looks like the manual page will need to be updated to account for `notifications` now being an object. Was there anything else you wanted to do with it before merging?

benjajaja:
@ulyssa thank you for fixing that! I think it's good to merge now.

I updated the manual page.

The barrage of past notifications is done, it seems like the SDK itself takes care of marking as delivered. I am still leaving in the line that discard notifications that are from before startup_time, but we could make that configurable later on.

PR #193, opened 2024-01-04
Support coloring entire message with the user color

### LINKED ISSUE #103: Different colours per users
Hi,

Is there a possibility to have different colours by users' text, so it would be easier to distinguish the text?

Many thanks

benjajaja:
Add a setting `message_user_color` that, if enabled, colors the message body in the chat just like the username. Some users like me prefer to have this visual cue, others may not like too much colors, so I think it's best to have an config option.

image

Fixes #103


ulyssa:
This is a great idea! Thank you for implementing this!

PR #194, opened 2024-01-06
fix empty vs. unsynced dms/rooms/spaces lists

### LINKED ISSUE #185: "Loading rooms..." over "You haven't joined any rooms yet" (same for DMs)
When loading my rooms, iamb displays a "You haven't joined any rooms yet". I'd say this message is appropriate for users who indeed have not joined any rooms yet, but not for the "loading" scenario.

Same for the "No direct messages yet!" message when loading DMs.

benjajaja:
Distinct empty-messages for some lists.
I'm not sure if using `client.sync_token().await.is_none()` is actually the correct way to this. It works when testing locally.

Fixes #185

### DIFF
--- a/src/base.rs
+++ b/src/base.rs
@@ -1105,7 +1105,7 @@ pub struct ChatStore {
     pub emojis: CompletionMap<String, &'static Emoji>,
 
     /// Information gathered by the background thread.
-    pub sync_info: SyncInfo,
+    pub sync_info: Option<SyncInfo>,
 
     /// Image preview "protocol" picker.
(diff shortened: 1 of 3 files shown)


ulyssa:
I've been thinking about this, and I'm pretty sure that starting the UI has to block on the initial sync and `refresh_rooms` finishing in order for restoring the layout or using a configured one to always work right and not race with the sync.  I've updated it to do so in #205 , so I think the message can stay as it is for now.

benjajaja:
Nice! Thank you!

PR #195, opened 2024-01-14
Add example of mapping "V" to toggle message selection mode

vodrazka:


ulyssa:
I like this idea! I actually use `V` a lot personally, so I'm not going to make this a default, but I've added a way in #217 to put custom mappings into the configuration. I've updated the PR to put the mapping in the example configuration so that people can opt-in to using `V` to toggle scrollback focus when they aren't don't use visual linewise.

PR #200, opened 2024-02-02
Update to matrix-sdk@0.7.1

ulyssa:
This is a WIP update to use a newer version of the matrix-sdk. Several notes:

- I still need a solution for migrating keys from the sled store into sqlite, so this only works on new logins
- This required bumping the MSRV to 1.70, so anyone still using 1.67 will need to update
- The sqlite store is now stored under `$XDG_DATA_HOME/iamb/profiles/$PROFILE/sqlite` instead of under `$XDG_CONFIG_HOME/iamb/profiles/$PROFILE/matrix` like the sled store was, as discussed in `#iamb-users:0x.badd.cafe` a couple months back

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,8 +22,8 @@ jobs:
       uses: actions/checkout@v3
       with:
         submodules: true
-    - name: Install Rust (1.67 w/ clippy)
-      uses: dtolnay/rust-toolchain@1.67
+    - name: Install Rust (1.70 w/ clippy)
+      uses: dtolnay/rust-toolchain@1.70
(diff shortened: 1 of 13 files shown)


ulyssa:
I cross-compiled some ARM binaries for gamingjones223 in `#iamb-users`. I'm posting them here in case anyone else wants to try them out:

- [iamb-armv7-linux.gz](https://github.com/ulyssa/iamb/files/14203362/iamb-armv7-linux.gz)
- [iamb-aarch64-linux.gz](https://github.com/ulyssa/iamb/files/14203415/iamb-aarch64-linux.gz)

PR #206, opened 2024-02-28
Fix Nix flake hashes

alin742:
This is probably related to #204 

I tried to install the package from the flake but I had an error regarding the hash of modalkit. When I removed it from flake.nix It complained about the hash of keybindings, so I generated a new hash and added it to the flake.nix and it was smooth sailing from there.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -24,9 +24,8 @@
           src = ./.;
           cargoLock = {
             lockFile = ./Cargo.lock;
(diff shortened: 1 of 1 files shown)


ulyssa:
Thanks for catching this! Yeah, Nix doesn't seem to like when the Git dependencies change. I'm hoping to update the `Cargo.toml` to use a crates.io dependency for modalkit soon, once I do another release.

PR #210, opened 2024-02-29
Add snap install instructions

popey:
I built and published an "unofficial" automated build of iamb in the snap store. This PR adds instructions to install the snap, and a badge which shows the current release in the (amd64) stable channel.


ulyssa:
Thank you for setting that up! :tada: I'll also add `snap` instructions to the website when I do the release.

PR #214, opened 2024-03-08
Nix flake updates

ErrorNoInternet:
- Updates the cargo hashes again
- Updated Rust  because matrix-sdk-common 0.7 was refusing to build on 1.67
- Used the current git commit hash as the version (appends -dirty if working tree is dirty)
- Small cleanups here and there

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,12 +1,15 @@
 {
   "nodes": {
     "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
       "locked": {
-        "lastModified": 1678901627,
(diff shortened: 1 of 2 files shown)


benjajaja:
Thank you!

Maybe we should add an action to build the flake some time.

ulyssa:
A GitHub action or some other way to test out building the Nix flake would be good! I've been thinking about trying to set up something on builds.sr.ht that runs after all of the GitHub actions have passed for testing out building on NetBSD and OpenBSD. If I get around to that, they [also have NixOS images](https://man.sr.ht/builds.sr.ht/compatibility.md), so that could also take care of validating that the `flake.nix` works.

PR #222, opened 2024-03-22
Support displaying shortcodes instead of emojis in messages

bbliem:
The font in my terminal doesn't support most emoji characters and shows ugly placeholders instead. For reactions, there is the setting `reaction_shortcode_display` that allows me to see human-readable shortcodes instead of the emojis. However, for emojis in messages there is no such feature.

I added a new setting `message_shortcode_display` that controls whether shortcodes are displayed instead of emojis in messages.

Replacing emojis with shortcodes happens in `TextPrinter` (or, for plain text, in `Message::show_msg()`).

I hope the changes are alright. I'd be happy to improve the code further and I'd be grateful for some feedback as I'm not very experienced in Rust yet.

### DIFF
--- a/docs/example_config.json
+++ b/docs/example_config.json
@@ -10,6 +10,7 @@
 	},
 	"settings": {
 		"log_level":                  "warn",
+		"message_shortcode_display":  false,
 		"reaction_display":           true,
 		"reaction_shortcode_display": false,
 		"read_receipt_send":          true,
--- a/docs/iamb.5.md
+++ b/docs/iamb.5.md
@@ -53,6 +53,10 @@ overridden as described in *PROFILES*.
 > Specifies the lowest log level that should be shown. Possible values
 > are: _trace_, _debug_, _info_, _warn_, and _error_.
 
+**message_shortcode_display** (type: boolean)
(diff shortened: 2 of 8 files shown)


ulyssa:
This is a good idea! Thank you for putting this together!

> I'd be happy to improve the code further and I'd be grateful for some feedback as I'm not very experienced in Rust yet.

It looks like you've already figured it out, but several things worth mentioning are:

- `cargo +nightly fmt` will take care of reformatting to the code
- `cargo clippy` will do linting and other checks; it's especially useful for getting simpler or improved alternatives to how you wrote something.
-  `cargo test` will run the test suite

The code looks good. One thing that I think would be good to add to the PR is to add tests to `test_emoji_shortcodes` of multi-codepoint Emojis, like:

-  🐻‍❄️ to `:snowflake:`  (bear + ZWJ + snowflake + variation selector 16)
- 🇨🇦 to `:canada:` (regional indicator letters C and A)

You're already using the `.graphemes` iterator, so I'm pretty sure testing those Emojis should just work without any other changes needed.

bbliem:
Thanks for the hints! And good idea about testing more complex emojis. Done!

PR #223, opened 2024-03-23
Support configuring user gutter width

mordquist:
I found the user gutter too wide by default, especially when using localpart or display name for usernames so i made it configurable. Seems to work fine but I'm not a rustacean so there is perhaps a more idiomatic way to do this.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -485,6 +485,7 @@ pub struct TunableValues {
     pub open_command: Option<Vec<String>>,
     pub notifications: Notifications,
     pub image_preview: Option<ImagePreviewValues>,
+    pub user_gutter_width: usize,
 }
 
 #[derive(Clone, Default, Deserialize)]
@@ -507,6 +508,7 @@ pub struct Tunables {
     pub open_command: Option<Vec<String>>,
     pub notifications: Option<Notifications>,
     pub image_preview: Option<ImagePreview>,
+    pub user_gutter_width: Option<usize>,
 }
 
 impl Tunables {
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -549,6 +552,7 @@ impl Tunables {
             open_command: self.open_command,
             notifications: self.notifications.unwrap_or_default(),
             image_preview: self.image_preview.map(ImagePreview::values),
+            user_gutter_width: self.user_gutter_width.unwrap_or(30 as usize),
```
ulyssa, line 555: `cargo clippy` suggests `30_usize` here, but I'm pretty sure you should just be able to do `30` and inference will figure it out.

src/message/mod.rs
```diff
@@ -576,10 +572,15 @@ impl<'a> MessageFormatter<'a> {
             text.lines.push(Line::from(vec![leading, date, trailing]));
         }
 
+        let user_gutter_empty = std::iter::repeat(' ')
+            .take(self.settings.tunables.user_gutter_width)
+            .collect::<String>();
+        let user_gutter_empty_span: Span<'a> = Span::raw(user_gutter_empty);
```
ulyssa, line 578: You can use the helper function from `util` here:

```suggestion
        let user_gutter_empty_span = space_span(self.settings.tunables.user_gutter_width, Style::default());
```

ulyssa (commented):
This looks good! I first chose 30 as the width before adding support for display names, so it's definitely a little wide with the other configurations.

There are a couple clippy warnings. I've commented on them and ways that I think you can improve on what it's recommending.

src/config.rs
mordquist: Had problems with `30` getting inferred as `u16` but I don't have that problem anymore. Maybe the LSP that was not working properly. Works as you suggested for me now.

mordquist:
Thank you for the quick review and suggestions.  Hopefully everything looks good now.

Issue #228, opened 2024-03-24
Multiple equal reactions from user are displayed multiple times


Kladki:
### Bug Description

When multiple `m.reaction` events are sent from a user with the same `key` and `rel_type` from the same user, it causes their reactions to add up rather than just being displayed once.  These are all from me:
![image](https://github.com/ulyssa/iamb/assets/158313939/e9feead2-0e78-490c-9c98-3ddac6696b19)

### Steps to Reproduce

1. Send a message
2. Get the event id
3. Use element devtools to send this event multiple times of type `m.reaction`:
```json
"m.relates_to": {
  "event_id": "${event_id}",
  "key": "👍",
  "rel_type": "m.annotation"
}
```
4. See issue

### Expected Behavior

Only one reaction of a certain emoji should be displayed for a single user.
> Multiple identical annotations (i.e., with the same event type and annotation key) from the same user (i.e., events with the same sender) should be treated as a single annotation.

https://spec.matrix.org/v1.10/client-server-api/#annotations-client-behaviour

### App Version

`0.0.8`

### Additional Context

Multiple reactions in the real world can be caused  by buggy clients, multiple clients reacting at the same time, lag, etc.

PR #239, opened 2024-03-29
 Fix reaction count when there are duplicate reaction events from a user

Kladki:
Closes https://github.com/ulyssa/iamb/issues/228


ulyssa:
Thank you for tracking #228 down and fixing it! :tada:  I hadn't realized that homeservers allow this, so handling this and #240 client side is definitely the right approach.

PR #253, opened 2024-04-10
Add `external_edit_file_suffix` to config

reynoldsme:
When using the `editor` command in iamb, the temporary file currently has no file extension. This is probably a reasonable default behavior, but some users may want to hint to their editor a particular file type such as `.md` or `.txt` to make use of linters, formatters, spell checkers, or other file type specific editor configurations.

This PR ~~preserves the default behavior~~ sets the default suffix to `.md` while adding a `external_edit_file_suffix` configuration option which accepts a string to append to the end of the temporary file name passed to the users editor.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -559,9 +559,11 @@ impl Tunables {
             notifications: self.notifications.or(other.notifications),
             image_preview: self.image_preview.or(other.image_preview),
             user_gutter_width: self.user_gutter_width.or(other.user_gutter_width),
-            external_edit_file_suffix: self
-                .external_edit_file_suffix
-                .or(other.external_edit_file_suffix),
+            external_edit_file_suffix: Some(
+                self.external_edit_file_suffix
+                    .or(other.external_edit_file_suffix)
+                    .unwrap_or_else(|| ".md".to_string()),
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -557,6 +559,9 @@ impl Tunables {
             notifications: self.notifications.or(other.notifications),
             image_preview: self.image_preview.or(other.image_preview),
             user_gutter_width: self.user_gutter_width.or(other.user_gutter_width),
+            external_edit_file_suffix: self
+                .external_edit_file_suffix
+                .or(other.external_edit_file_suffix),
```
ulyssa, line 564: I would be fine with an `.unwrap_or_else(|| ".md".into())` here. iamb interprets the text as markdown, so that's probably the most useful default.

ulyssa:
Merged! Thank you for thinking to add this! :tada: Getting syntax highlighting is I imagine one of people's big reasons to use an external editor, so this should make that much easier on them. (Although it would be nice to eventually implement it in the message bar, too.)

reynoldsme:
Happy to help, and thank you for creating and maintaining iamb!

PR #257, opened 2024-04-13
Fix image preview placement when messages are preceded by a date in the timeline

reynoldsme:
Currently image previews are printed one line too high if the message had a date inserted into it indicating a change of date between the current and preceding message. With this PR we conditionally check if a date was inserted and adjust the y offset used to place the rendered the image.

Tested in: `XTerm(374)`, `wezterm 20240203-110809-5046fc22`, and `konsole 24.02.1`.

Before change:
before change

After change:
after

### DIFF
--- a/src/message/mod.rs
+++ b/src/message/mod.rs
@@ -961,6 +961,8 @@ impl Message {
         let proto = proto.map(|p| {
             let y_off = text.lines.len() as u16;
             let x_off = fmt.cols.user_gutter_width(settings);
+            // Adjust y_off by 1 if a date was printed before the message to account for the extra line.
+            let y_off = if fmt.date.is_some() { y_off + 1 } else { y_off };
             (p, x_off, y_off)
         });
 


src/message/mod.rs
```diff
@@ -961,6 +961,8 @@ impl Message {
         let proto = proto.map(|p| {
             let y_off = text.lines.len() as u16;
             let x_off = fmt.cols.user_gutter_width(settings);
+            // Adjust y_off by 1 if a date was printed before the message to account for the extra line.
+            let y_off = if fmt.date.is_some() { y_off + 1 } else { y_off };
```
ulyssa, line 965: Good catch! This isn't quite right though, since if there's a reply then `text.lines` will already include the date, and we'll double count it here. The date printing code should probably be moved out of `MessageFormatter::push_spans` and into its own `MessageFormatter::push_date` method that gets called before pushing the reply text instead.

ulyssa:
Merged! Thank you for fixing this!

Issue #246, opened 2024-04-02
Empty :editor file sends blank line


saygo-png:
Opening an editor with `:editor` (neovim for me) then quiting it without making any changes sends an empty line as a message. Ideally it would only send something if the temp file isn't empty.

saygo-png:
feature was merged

saygo-png:
feature was merged

PR #275, opened 2024-04-27
Trim `:editor` output and check if it's empty

Andrew-Collins:
Hi,

Quick fix for #246 by trimming the editor output and checking if it's empty.


ulyssa:
Thank you for fixing this!

Issue #268, opened 2024-04-16
:logout fails


landryb:
using 0.0.9 on OpenBSD, `:logout` command from #162 exists (eg `:log<tab>` completes it) but it fails with `Failed command: Invalid argument`

mordquist:
You need to specify the account you want to log out of to avoid accidentally logging out of the wrong account. Try logging out with `:logout @user:example.com`.

PR #277, opened 2024-04-27
Add error for missing username on `:logout`

mordquist:
Adds error message if no argument is supplied.
Now `Failed command: Error: Missing username` is displayed if no username is given.

Did not remove username requirement for single account clients since different behavior depending on number of accounts could be confusing.

Fix for #268


ulyssa:
Thank you for fixing this! I should probably update the invalid argument variant in the next modalkit release to take a `String` with details.

Issue #50, opened 2023-03-06
Support adding new aliases to a room


ulyssa:
Rooms can be given an alias when created via `:create ++alias=localpart`, but there's not a way to add them to existing rooms, or to modify which alias is the canonical alias.

ulyssa:
Fixed by @Lymkwi in #279.

PR #279, opened 2024-05-01
Implement set/unset/show for alternative and canonical aliases

Lymkwi:
This commit solves #50 by providing a way to:
 - Set the canonical alias of a room (destroying the previous one if any)
 - Unset the canonical alias of a room (destroying the previous one if any)
 - Add an alternative alias to a room
 - Delete an alternative alias to a room (destroying it afterwards)


ulyssa:
This is a really great addition! Thank you so much for adding this, @Lymkwi ! :partying_face: 

While playing around with this, I ended up changing the behaviour when setting a new canonical alias to put it into the alt aliases instead, so that people have a natural way to set new canonical aliases while keeping the old around for joining. If someone does want to get rid of it, then they can do `:room alias unset #oldcanonicalalias:homeserver.com`.

Issue #276, opened 2024-04-27
[Feature request] Send notifications depending on terminal focus


hardcore-sushi:
Currently, notifications are not sent if the room is in foreground on the terminal. However, users may still want to receive these notifications in the case their terminal is not visible, for example when using window workspaces or scratchpads. It looks like iamb already knows whether the terminal is focus via the `Application::focused` attribute. Couldn't that be used to decide whether to send a desktop notification?

ulyssa:
Implemented in #281 by @Andrew-Collins

PR #281, opened 2024-05-04
Allow notifications on open room if terminal not focused

Andrew-Collins:
Hi,

This PR is to implement #276.

It does this by putting the `focused` state of the application into the store and then checks it in the notifications handler.

This results in notifications being muted only if the room is currently displayed and the terminal is in focus.

If you would like a config entry to control this just let me know. :)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -355,9 +355,13 @@ impl Application {
                     // Do nothing for now.
                 },
                 Event::FocusGained => {
+                    let mut store = self.store.lock().await;
+                    store.application.focused = true;
                     self.focused = true;
                 },
                 Event::FocusLost => {
+                    let mut store = self.store.lock().await;
+                    store.application.focused = false;
                     self.focused = false;
                 },
                 Event::Resize(_, _) => {
--- a/src/base.rs
+++ b/src/base.rs
@@ -1344,7 +1344,7 @@ impl ChatStore {
             sync_info: Default::default(),
             draw_curr: None,
             ring_bell: false,
-            focused: false,
+            focused: true,
         }
     }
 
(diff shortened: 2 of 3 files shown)


benjajaja:
Looks great, thank you!

src/main.rs
```diff
@@ -355,9 +355,13 @@ impl Application {
                     // Do nothing for now.
                 },
                 Event::FocusGained => {
+                    let mut store = self.store.lock().await;
+                    store.application.focused = true;
```
benjajaja, line 359: Do we always get this event on startup?

Andrew-Collins: Sorry busy week.
Thanks for the catch @benjajaja, I tested and we do not get the event on startup.
This has been resolved by setting the default of `focused` in the store to `true`.

benjajaja: Thank you! There is an edge where iamb is started without focus, but I don't think it's possible to query the focus state from the terminal, and it's not a big deal.

benjajaja (approved):

src/main.rs
Andrew-Collins: Np :)

@benjajaja, I was also thinking about that edge case this morning, but I wasn't able to think of an easy way to resolve it either...

The only feasible way I can think of is to execute `crossterm` on entry to main, and run login/sync in the background.

That would end up being a fairly significant change, but if you are interested I can try and tackle it in another PR?

benjajaja: I don't understand how that would give us the focus-state of the terminal. But I'm not really worried about that edge case, this PR is much more important. Thanks again!

ulyssa:
@Andrew-Collins This is a great addition, thank you! :pray:

And I believe that @benjajaja is correct that we don't get the focused state on startup, but only whenever the window changes. (And that doesn't always happen with every terminal either). Defaulting `focused` to `true` is the right move. :+1:

PR #283, opened 2024-05-11
Fix openSUSE link and installation command in README

photosheep:
The package is now hosted in the main openSUSE Tumbleweed repositories. This corrects the README to use the openSUSE repo instead of a user-maintained one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,6 @@
 
 </div>
 
-
 ## About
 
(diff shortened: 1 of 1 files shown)


ulyssa:
Thank you for updating the instructions! I'll update the website too to reflect it now being in the main repos. :tada:

PR #286, opened 2024-05-14
Add missing darwin build dependency

voidcontext:


ulyssa:
I think this probably changed when adding the `notify-rust` crate. Thank you for fixing it!

PR #300, opened 2024-06-17
Fix a compiler warning

jarkkojs:
```
warning: field `1` is never read
   --> src/message/html.rs:263:30
```


ulyssa:
Merged as part of #301 .

PR #301, opened 2024-06-17
Fix newer Clippy warnings for 1.80

jarkkojs:
I'd like to take this client to my own use but as long as it breaks "broken glass" invariant I really cannot do that ;-) [1]

Is there something broken in the CI, I see that there are some clippy bits but cannot comprehend why those checks are not working?

[1] "Fight software rot" chapter in "Pragmatic Programmer"

### DIFF
--- a/src/base.rs
+++ b/src/base.rs
@@ -1346,9 +1346,7 @@ impl ChatStore {
 
     /// Get a joined room.
     pub fn get_joined_room(&self, room_id: &RoomId) -> Option<MatrixRoom> {
-        let Some(room) = self.worker.client.get_room(room_id) else {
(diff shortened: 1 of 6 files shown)


ulyssa:
Thank you for cleaning this up!

> Is there something broken in the CI, I see that there are some clippy bits but cannot comprehend why those checks are not working?

Each new edition of Rust/Clippy stablizes or adds new lints. Since I don't want to force folks to have to be responsible for fixing new lints every time there's a new release of Rust, the CI workflow runs Clippy using the MSRV for iamb (currently 1.70) which doesn't warn on the checks fixed here.

PR #302, opened 2024-06-17
Fix `cargo fmt --check` issues

jarkkojs:


ulyssa:
It looks like you're running `cargo fmt`  instead of `cargo +nightly fmt`. (You should be seeing warnings printed from `rustfmt` when you do so, listing the unstable features that it's running without.)

PR #303, opened 2024-06-18
Add Hombrew as install method on MacOS

Backroom8816:
Hi,

The aim of it is to update iamb's documentation to include the Hombrew repository link and the command to install iamb on MacOS.

Also, I alphabetically sorted the install instructions for each platform.


ulyssa:
Thank you for updating the docs and alphabetizing the install instructions! It's wonderful to see so many platforms in the README! I'll add Homebrew instructions to the site, too.

Issue #304, opened 2024-06-18
Per Room Notification Settings?


rmsthebest:
Hi,
Is it possible to configure the notification options on a per room basis in iamb?

I have some rooms that are important to me, like family chats, and some less important public/group chats I don't want notifications for.

Edit: Looking at the code it looks like iamb already has `RoomNotificationMode`, but only on the notification trigger side. With no way to configuring them. Perhaps it could be added as a room command, "room notification set `<notification level>`"

benjajaja:
In the meantime, it should be possible to configure those settings in another client (e.g. element.io webapp).

ulyssa:
Fixed in #305

PR #305, opened 2024-06-20
Add command to set per-room notification levels

rmsthebest:
As mentioned in the issue #304, I couldn't find a way to set notification options.

Here is an attempt to implement it. It isn't working, I'm not sure why so help is appreciated.

I apologize for breaking formatting, maybe my editor isnt picking up the rustfmt.toml from the repo?
If we get this working we can fix that later

### DIFF
--- a/docs/iamb.1
+++ b/docs/iamb.1
@@ -122,6 +122,25 @@ View a list of members of the currently focused room.
 Set the name of the currently focused room.
 .It Sy ":room name unset"
 Unset the name of the currently focused room.
+.It Sy ":room notify set [level]"
+Set a notification level for the currently focused room.
+Valid levels are
+.Dq mute ,
+.Dq mentions ,
+.Dq keywords ,
+and
+.Dq all .
+Note that
+.Dq mentions
+and
+.Dq keywords
+are aliases for the same behaviour.
+.It Sy ":room notify unset"
+Unset any room-level notification configuration.
+.It Sy ":room notify show"
+Show the current room-level notification configuration.
+If the room is using the account-level default, then this will print
+.Dq default .
 .It Sy ":room tag set [tag]"
 Add a tag to the currently focused room.
 .It Sy ":room tag unset [tag]"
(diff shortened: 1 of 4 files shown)


ulyssa (commented):
This looks good! Thank you for implementing this! 

> I apologize for breaking formatting, maybe my editor isnt picking up the rustfmt.toml from the repo?

The formatting thing is probably because your editor isn't using `cargo +nightly fmt` to enable the nightly-only options that are in `.rustfmt.toml`. I've run it by hand and pushed to the branch to make it easier to read.

> If we get this working we can fix that later

What part wasn't working for you? I tried playing around with this locally and it seemed to be working at a first glance to me. I could mute and unmute rooms and messages sent to that room by other accounts disappeared/appeared based on the settings. The settings changes are also visible in Element.

I've added an `unset` command that uses `delete_user_defined_room_rules()` to restore settings to the account default, and I've removed some of the level name variants to keep it a small easy-to-type list for now.

ulyssa:
I've fixed the merge conflicts and added support for using with the `RoomAction::Show` action that was added in #279.

ulyssa:
From what I can tell playing around with this, it seems good to me! If there are any other issues we can fix those in a follow-up PR.

rmsthebest:
> This looks good! Thank you for implementing this!
> 
> > I apologize for breaking formatting, maybe my editor isnt picking up the rustfmt.toml from the repo?
> 
> The formatting thing is probably because your editor isn't using `cargo +nightly fmt` to enable the nightly-only options that are in `.rustfmt.toml`. I've run it by hand and pushed to the branch to make it easier to read.
> 
> > If we get this working we can fix that later
> 
> What part wasn't working for you? I tried playing around with this locally and it seemed to be working at a first glance to me. I could mute and unmute rooms and messages sent to that room by other accounts disappeared/appeared based on the settings. The settings changes are also visible in Element.
> 
> I've added an `unset` command that uses `delete_user_defined_room_rules()` to restore settings to the account default, and I've removed some of the level name variants to keep it a small easy-to-type list for now.

So a few days after I wrote this I noticed that the notification settings seems to had changed, because it changed them on my phone, but in iamb on my desktop it didnt seem to care about the per room settings at all.

I'm glad it works for you though.
Thanks for fixing it up and merging!

PR #329, opened 2024-08-18
Handle message marks on non-64-bit platforms

ulyssa:
This fixes the issue with running the iamb tests on i686 seen in void-linux/void-packages#49652. I initially tried to do this in modalkit, but ended up deciding that `usize` is really the right type for `Cursor` and not `u64`. I've instead done the conversion here by using `usize::MAX` to make the hash predictably convertible.

### DIFF
--- a/src/message/mod.rs
+++ b/src/message/mod.rs
@@ -4,7 +4,7 @@ use std::cmp::{Ord, Ordering, PartialOrd};
 use std::collections::hash_map::DefaultHasher;
 use std::collections::hash_set;
 use std::collections::BTreeMap;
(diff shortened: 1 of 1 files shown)


ulyssa:
Also, writing down how to run the tests w/ i686 in case I need to do it again in the future:

```
% sudo apt-get install crossbuild-essential-i386
% CC_i686_unknown_linux_gnu=i686-linux-gnu-gcc CARGO_TARGET_I686_UNKNOWN_LINUX_GNU_LINKER=i686-linux-gnu-gcc \
    cargo test --target i686-unknown-linux-gnu
```

PR #343, opened 2024-08-30
fix: unused result of `format!`

Nemo157:
This shows a warning in 1.81+, I assume it was meant to be printed to stdout along with the rest of the messages in this function.


ulyssa:
Yes, that is meant to be a `prinln!`! Thank you for fixing this!

PR #344, opened 2024-08-30
Support enabling multiple notification sinks

### LINKED ISSUE #340: Allow enabling both desktop and bell notifications
I would like to use both these notification channels for different purposes:

 * desktop notifications to have a transient preview of messages while I'm at my machine in case I want to respond quickly
 * bell notifications to put a marker on the tmux window iamb is inside while I'm away and the desktop notification will have already disappeared

(the latter is really a better version of what I was trying to use `monitor-activity` in #339 for, since it will also handle activity in hidden rooms)

Nemo157:
Fixes #340, as I mention there I have a usecase for using both kinds of notifications at the same time.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -398,23 +398,70 @@ pub enum UserDisplayStyle {
     DisplayName,
 }
 
-#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq)]
-#[serde(rename_all = "lowercase")]
-pub enum NotifyVia {
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct NotifyVia {
     /// Deliver notifications via terminal bell.
-    Bell,
+    pub bell: bool,
     /// Deliver notifications via desktop mechanism.
     #[cfg(feature = "desktop")]
-    Desktop,
+    pub desktop: bool,
 }
+pub struct NotifyViaVisitor;
 
 impl Default for NotifyVia {
     fn default() -> Self {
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -398,14 +398,17 @@ pub enum UserDisplayStyle {
...
 #[serde(rename_all = "lowercase")]
 pub enum NotifyVia {
     /// Deliver notifications via terminal bell.
     Bell,
     /// Deliver notifications via desktop mechanism.
     #[cfg(feature = "desktop")]
     Desktop,
+
+    #[serde(untagged)]
+    Multiple(Vec<Self>),
```
Nemo157, line 411: This technically allows `[["bell", "bell"], "desktop"]` and other silly settings, but for code simplicity that feels fine to me. The other approach would be to use something like `bitflags` but I'm not sure how complicated deserializing that would be.

ulyssa: I think that making this look like a bitwise-or would be good so that folks can write `via: "bell|desktop"`. There's an example of what that looks like [here for VimModes](https://github.com/ulyssa/iamb/blob/3355eb2d2622f28d2fed6eb00a6aa78868a87b83/src/config.rs#L181-L223) which you could copy for making a `NotifyViaSet(Vec<NotifyVia>)` type.

Nemo157: Finally got around to changing it to look like a bitwise-or, and added parser tests for it.

ulyssa:
This looks great! Thank you for adding tests for it, too!

Issue #345, opened 2024-08-30
sixel rendering issues in wezterm and tmux


Nemo157:
I'm really not sure where the issue is, I spent a while trying to come up with something more useful to open an issue about, but I couldn't track down anything really useful.

When using sixel preview rendering it fails to display when running in wezterm directly, or when running in tmux within both wezterm and foot. In wezterm, with or without tmux, it mostly seems to render previous lines leftover from scrolling where the image should be. In tmux+foot it mostly seems to be blank, but e.g. when scrolling messages there's a single frame flicker of the image.

It does seem to work correctly when running directly in foot.

I also use sixel rendering in vifm which generally works in all the above situations, and running the `ratatui-image` example seems to work too. Just in case it was related to the `StatefulProtocol` which it says is more reliable I ported iamb over to use that, it didn't seem to make any difference.

reynoldsme:
@Nemo157 did you try directly [setting the image preview protocol](https://iamb.chat/configure.html#image-previews) in your iamb config, just in case the ratatui-image examples and iamb are using different image protocol detection logic?

For me running an iamb debug build of latest main commit and WezTerm-20240203-110809-5046fc22-Ubuntu20.04.AppImage, images show up as expected, but a copy of the first line of some images will sometimes appear and disappear at column `0` of the line on which the image occurs. If this is inconsistent between platforms or even just different machines, it makes me suspect the issue may be on the Wezterm side.

<img width="615" alt="wezterm-ratatui-image" src="https://github.com/user-attachments/assets/9bdc2dbf-9e52-4813-91e2-9d4e52c3be0e">

 The ratatui-image readme does imply there are known bugs when using ratatui-image in sixel mode in Wezterm https://github.com/benjajaja/ratatui-image?tab=readme-ov-file#compatibility-matrix , but I don't know if this specific issue is something that is already being tracked on not. There is an open (but locked) issue for sixel on the Wezterm side, and I see the ratatui-image author has commented on that issue in the past. https://github.com/wez/wezterm/issues/217

Nemo157:
Yes, I've always been hardcoding the protocol in the config. I think I have noticed that specific bug too, but less often than issues with the rendering in the frame the image is meant to be. I wouldn't be surprised if this is all bugs in the TEs, I'm just a bit unsure since I didn't notice the same issues with other sixel apps.

[...]

PR #352, opened 2024-09-13
Fix `ratatui-image` tmux detection when used with a configured image protocol

Nemo157:
I finally tracked down part of the problems from #345, `ratatui-image` only does its tmux detection when `guess_protocol` is called.

This is required for kitty (and presumably iterm2) protocol to work inside tmux, for sixel it's not _required_ since tmux actually implements sixel support itself, but it does appear to give better results by bypassing tmux and passing the sixel data directly to the terminal.

### DIFF
--- a/src/base.rs
+++ b/src/base.rs
@@ -1276,10 +1276,12 @@ fn picker_from_termios(protocol_type: Option<ProtocolType>) -> Option<Picker> {
         },
     };
 
+    // `guess_protocol` also does tmux detection,
+    // run it always then overwrite the guessed protocol if needed
+    picker.guess_protocol();
+
     if let Some(protocol_type) = protocol_type {
         picker.protocol_type = protocol_type;
-    } else {
-        picker.guess_protocol();
     }
 
     Some(picker)


benjajaja:
Thank you!

reynoldsme:
I will agree that with this patch, sixel in tmux is *significantly* more usable. Images are consistently visible and the performance issues appear to be gone. The only downside I can see is that it looks as if tmux is no longer tracking the placement of the sixels internally and as a result, terminal layout starts to break as soon as you add tmux splits as sixels can end up getting placed wherever the cursor is. 

sixel-tmux-bypass2

In particular, if a sixel gets renders past the current bottom of the terminal, the terminal itself will scroll instead of tmux, and that breaks layout pretty badly (Note how the tmux status line is moving up the screen on every sixel refresh in the next screenshot).

tmux-sixel-bypass3

Nemo157:
Right, that makes a lot of sense, and actually seems to happen with iTerm2 as well, since the cursor is not under the apps control when it's not the active pane and so since the passthrough'd escape sequences are cursor dependent they render in the wrong place. Kitty seems to be the only protocol that actually renders correctly in that situation since its escape sequence isn't related to placement, the placement happens purely through the unicode characters.

ulyssa:
@Nemo157 Thank you for looking into this and figuring out the fix, and thank you to @benjajaja for reviewing it!

It sounds like from @reynoldsme that this is a net improvement when running in `tmux` so I'm going to go ahead and merge it. There is code that should prevent drawing images past the bottom of the window, but if I understand right it sounds like the image doesn't get cleared when the window gets resized during the `tmux` split. That will probably require a follow-up change to improve the situation.

JIbald:
great job, thank you.

Issue #362, opened 2024-10-10
Image preview not working in threads


StHottus:
Image preview is working perfectly in rooms and dms, but in threads, it's just not showing up.

Using iamb 0.0.10 on kitty 0.36.4 with 
```
[settings]
image_preview = {}

```
![Bildschirmfoto_2024-10-10_15-58-41](https://github.com/user-attachments/assets/e4e92846-2986-43ac-8523-7613f1f45854)
![Bildschirmfoto_2024-10-10_15-59-04](https://github.com/user-attachments/assets/6ace881d-20fd-4e0f-90cb-722bdc2e7832)

PR #366, opened 2024-10-16
Fix image preview in replies and threads

Andrew-Collins:
Hi,

This PR is a fix for #362

The root cause was that the image protocol from a reply header was being ignored.

An image in a reply header is now handled in the same way as an image in the body of the message, with the image position offsets modified to account for the difference in formatting (vertical lines, indent, etc).

It was necessary to break out `Message::show_msg` from `MessageFormatter::push_in_reply` due to the lifetime of the image protocol. 

`Message::show_with_preview` now returns the image protocols of both the body and the reply header. `Scrollback::render` was updated to handle this.

### DIFF
--- a/src/message/mod.rs
+++ b/src/message/mod.rs
@@ -712,14 +712,13 @@ impl<'a> MessageFormatter<'a> {
     fn push_in_reply(
         &mut self,
         msg: &'a Message,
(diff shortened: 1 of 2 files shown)


ulyssa:
Thank you for tracking this bug down! :pray:

PR #374, opened 2024-12-03
Mention Gentoo's GURU ebuild in the README

repomansez:


ulyssa:
Thank you @repomansez ! I've made an equivalent PR that will update the website at ulyssa/iamb.chat#33 too.

Issue #379, opened 2024-12-08
cannot show room topic


daef:
according to the [docs](https://iamb.chat/commands.html?highlight=topic%20show#iamb-commands) i should be able to show the current topic using `:room topic show`, but when i try i just get "Failed command: Invalid argument". skimming [commands.rs](https://github.com/ulyssa/iamb/blob/main/src/commands.rs#L471) i don't see where `topic show` is supposed to be handled either. is the functionality simply missing or am i missing something? 

edit: funny thing: `RoomAction::Show(RoomField::Topic)` seems to be [implemented](https://github.com/ulyssa/iamb/blob/main/src/windows/room/mod.rs#L583)

PR #380, opened 2024-12-08
Support `:room topic show`

daef:
(i hope) this fixes https://github.com/ulyssa/iamb/issues/379 ...


ulyssa:
Thank you for catching this! :pray: It looks like #279 added what was needed for this, but missed the subcommand parsing part.

PR #396, opened 2025-02-20
Updated rust version and added sqlite in flake.nix

oddlid:
The original flake.nix did not work to build the latest commit with. This one does, at least in my local tests. I've only tested on an MBP M1, and I don't have any other system with nix (yet), so please test if anyone have other platforms.


ulyssa:
Thank you for bumping this! :pray: The minimum Rust version changed for #386, and I suspect that's why the nix flake broke.

Issue #190, opened 2023-12-28
Feat: nested spaces


sajattack:
Forgive me if this has already been mentioned but I did a quick look through your issue board and didn't see it.

Element supports spaces within spaces - which comes in handy especially if you're bridging discords to matrix.

It would be nice if these nested spaces could be viewed with proper hierarchy.

Allow me to illustrate:

![2023-12-28-103125_613x665_scrot](https://github.com/ulyssa/iamb/assets/1562711/2c81a169-b1de-4802-84f6-2f24d836d06f)
![2023-12-28-103226_363x521_scrot](https://github.com/ulyssa/iamb/assets/1562711/8d38647d-1a13-4178-a944-995dd81f7794)

Ramblurr:
Nested spaces can be navigated with hjkl + enter, but there's no easy way to go back up a level, which would be handy. 

Alternatively I guess a tree view would be nicer (though more complicated to implement).

VAWVAW:
Since a space doesn't always have an unambiguous parent there isn't always a simple answer as to what "up" should be. You could define "up" as "where I came from" but that can already be achieved using the jump list (shortcut `<C-O>`).

PR #405, opened 2025-03-25
Navigate nested spaces

VAWVAW:
addresses #190

This adds the command `:spaces!` to open a space list that only shows spaces which are not contained in other joined spaces. This needs to fetch the members of every joined space when it opens and hangs for multiple seconds with like 20 joined spaces. So draft for now.


VAWVAW:
This approach isn't a good way to implement this.

PR #422, opened 2025-05-16
Update to latest `modalkit`, `modalkit-ratatui` and `ratatui-image`

ulyssa:
This updates to `modalkit@0.0.21`, `modalkit-ratatui@0.0.21`, and `ratatui-image@5.0.0` (but pointing at a Git SHA that includes benjajaja/ratatui-image#91 for now).

### DIFF
--- a/src/base.rs
+++ b/src/base.rs
@@ -1319,7 +1319,7 @@ fn emoji_map() -> CompletionMap<String, &'static Emoji> {
 
 #[cfg(unix)]
 fn picker_from_termios(protocol_type: Option<ProtocolType>) -> Option<Picker> {
-    let mut picker = match Picker::from_termios() {
+    let mut picker = match Picker::from_query_stdio() {
         Ok(picker) => picker,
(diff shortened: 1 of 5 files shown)


src/base.rs
```diff
@@ -1319,20 +1319,16 @@ fn emoji_map() -> CompletionMap<String, &'static Emoji> {
...
         Ok(picker) => picker,
         Err(e) => {
             tracing::error!("Failed to setup image previews: {e}");
             return None;
         },
     };
 
-    // `guess_protocol` also does tmux detection,
-    // run it always then overwrite the guessed protocol if needed
-    picker.guess_protocol();
```
ulyssa, line 1332: @benjajaja To make sure I'm understanding right, `Picker::from_query_stdio` does the same thing as `from_termios` plus `guess_protocol`, so there's nothing that should replace this, right?

benjajaja: Yes, that's correct.

PR #430, opened 2025-05-29
Notification improvements

VAWVAW:
This sets the urgency for all notifications to normal and truncates the body at 5000 instead of 100.

Some notification servers that let you set options depending on urgency have no option to match "no urgency specified".

A lot of messages are longer than 100 characters and most notification servers will truncate the message to a limit they can still display.


ulyssa:
Makes sense! Thank you for improving this! :pray:

PR #433, opened 2025-05-30
Fix most incorrect unreads on startup

VAWVAW:
Read receipts can either be by thread or unthreaded (for the whole room). Iamb only sends unthreaded receipts but would load all on initial sync with the latest specified in the list taking precedence. It would also request receipts for the main thread and not unthreaded.

This has reduced the number of rooms that get incorrectly marked as unread on startup by at least 90% for me.


ulyssa:
Thank you for hunting this down! This seems like it's probably the right fix for the time being. In the long run, I guess iamb should be sending both `ReceiptType::Main` and `ReceiptType::Threaded`, but that'll require some additional changes to `RoomInfo::{event_receipts,user_receipts}` to make sure the main timeline and the threads are tracked independently.

The rest of the unreads are probably due to not having state events in the timeline. I have a local change I've been working on to insert most of them into the scrollback, so hopefully that will help clear up the other 10% for you and other folks.

VAWVAW:
Actually never mind. The problem seems to be some kind of race condition because I still have a lot of rooms marked as unread on most restarts.

But this PR fixed a bug anyways.

youwen5:
> Actually never mind. The problem seems to be some kind of race condition because I still have a lot of rooms marked as unread on most restarts.
> 
> But this PR fixed a bug anyways.

Yeah I'm seeing most rooms marked unread on restart. Curiously I never had this issue on the matrix.org homeserver and initially it was working fine after I switched to a self-hosted tuwunel server. Not sure what started causing it.

VAWVAW:
Could you try whether #451 fixes your problem?

youwen5:
Seems to have had no effect.

VAWVAW:
Could you try https://github.com/VAWVAW/iamb/tree/debug-unreads, add the room id and name of a room you want to debug in `src/main.rs:39` and send me the logs? Debugging these read receipts is somewhat tricky since it is influenced by the homeserver.

PR #435, opened 2025-05-30
Remove useless code wrapping `load_event_receipts`

VAWVAW:
`MatrixRoom::load_event_receipts` is just a wrapper around [`StateStore::get_event_room_receipt_events`](https://docs.rs/matrix-sdk-base/latest/matrix_sdk_base/store/trait.StateStore.html#tymethod.get_event_room_receipt_events) which loads the receipts from the memory or sqlite store.

According to the [spec](https://spec.matrix.org/v1.14/client-server-api/#receipts) there is no way to explicitly request read receipts from the server.

The initial sync sends all receipts for the room and the incremental sync gets all updates both of which get handled in the [event handler](https://github.com/ulyssa/iamb/blob/f39261ff84678feae1e58a5e050fbe4bd4ae27b5/src/worker.rs#L1033).

### DIFF
--- a/src/worker.rs
+++ b/src/worker.rs
@@ -70,7 +70,6 @@ use matrix_sdk::{
         room::RoomType,
         serde::Raw,
         EventEncryptionAlgorithm,
-        EventId,
         OwnedEventId,
         OwnedRoomId,
         OwnedRoomOrAliasId,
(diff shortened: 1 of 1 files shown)


VAWVAW:
Another thing I noticed is that the spec specifies
> Clients SHOULD NOT send read receipts for events sent by their own user.
which is something iamb does.

VAWVAW:
For some reason the handler doesn't always get called so leaving this in.

Issue #292, opened 2024-05-25
Rooms get incorrectly marked as unread on startup


immae:
When I start iamb (version 0.0.9), multiple rooms (both "public" and "DM", even some DM rooms where the other part has left ages ago) get marked as unread, even though they don’t contain any new message. Other clients don’t show any "unread" marker either for those rooms.

I cannot find any common factor for those unread markers: in some of them I’m the last one to speak, but not all of them. The "read" marker with my name is at the last message, some rooms has no reaction or event that could have landed inbetween.

However the list of "unread" rooms seems to be consistent : if I mark everything as read, quit and rerun iam, then the exact(-ish?) same list of room will get marked as unread.

I don’t know what to do to debug, if I can be of any help to figure out what’s wrong

mordquist:
I've got the same experience as you and have started looking at it but don't have a complete picture of the problem yet. The function responsible for marking rooms as unread is located [here](https://github.com/ulyssa/iamb/blob/497be7f0998d850e123362aa0ac2f8fdb8362066/src/base.rs#L965).

I have noticed problems with incorrectly marked rooms when:
- the latest message in a room is part of a thread (other than the main thread)
- the latest event in a room is not a message (e.g. someone joined after the last message)

Other possible issues I've not looked in to are threaded vs unthreaded and private receipts.

freswa:
I've got the same behavior and a lot of warnings in the logs that read like this:
```
2024-07-31T10:39:56.591601Z  WARN iamb::worker: Failed to set read receipt: the server returned an error: [403 / M_FORBIDDEN] You don't have permission to access that event. room_id="!xxxxxxxxxxxxx:matrix.org"
```

mcepl:
I can see the same, and it looks to me like iamb somehow doesn’t push to the server changes in the statuses.

Blub:
Still happening here as well, but I don't see any such read-receipt errors in the logs.

timocontecht:
I confirm the issue

VAWVAW:
This is fixed for me except when the last message was in a thread. But I am working on that.

(I also had some rooms with weird states where I just needed to send a new message.)

Blub:
It's better now (git main from today) but still happening...

oddlid:
I just compiled from main, and it seems to depend on the server I connect to. 
[...]

PR #438, opened 2025-05-31
Implement receipts per thread

VAWVAW:
This saves read receipts per message thread and fixes the remaining issues from #292.

remaining issues to be fixed in later PRs:

- unread messages in threads are not marked as "Unread" in the room list
  - previously when the last message in the room was in the thread the room was always unread
  - there is way to discover which thread in the room is unread
  - should be implemented with a `:threads` overview
- users that have both an unthreaded receipt and a receipt for the main thread are shown twice in the receipt gutter
  - this is only an issue if some users in the room use clients that only implement unthreaded receipts like old versions of iamb
- `load_older_one` only loads receipts for the main thread

includes #435

### DIFF
--- a/src/base.rs
+++ b/src/base.rs
@@ -12,6 +12,7 @@ use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 use emojis::Emoji;
+use matrix_sdk::ruma::events::receipt::ReceiptThread;
 use ratatui::{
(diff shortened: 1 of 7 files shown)


ulyssa:
This is great work! :sparkles: Thank you so much for implementing this!

PR #441, opened 2025-05-31
Document every iamb command in the manpage

VAWVAW:


ulyssa:
Thank you for noticing that these were missing and adding them! :pray:

PR #445, opened 2025-06-04
Show newest edit of messages even after application restart

### LINKED ISSUE #431: Old edits are not shown
# steps to reproduce

1. send a message in any room
2. edit that message
3. restart iamb
4. view the same message

After 2. the changed message is shown. After 3. the original message is shown.

VAWVAW:
Save all revisions of an edited message. This allows us to show the newest version, even if they are loaded out of order or an edit is redacted.

In case the edit event is loaded before the message itself, it is saved to `RoomInfo::unloaded_edits`.

## possible future additions

- The edit history could be exposed to the user with a command like `:edits`, making it possible to view and redact edit events.

fixes #431

### DIFF
--- a/src/base.rs
+++ b/src/base.rs
@@ -3,7 +3,7 @@
 //! The types defined here get used throughout iamb.
 use std::borrow::Cow;
 use std::collections::hash_map::IntoIter;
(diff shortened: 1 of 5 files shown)


ulyssa:
Thank you for figuring this out! :pray:

Issue #444, opened 2025-06-02
Preview for redacted images is not removed


VAWVAW:
Steps to reproduce

- receive an image for which a preview is rendered
- receive a redaction event for the image

The timline renders first the image preview and then the text "[Redacted]" in the next line. I would expect the image preview to be deleted as well.

PR #448, opened 2025-06-05
Remove image preview on message redact

VAWVAW:


ulyssa:
Thank you for fixing #444!

VAWVAW:
I just noticed that if the image is downloading while the redaction is received, it will still show the placeholder saying "Downloading..." with the "[Redacted]" below it. This is probably a race condition in the communication with the worker and not worth hunting down.

PR #451, opened 2025-06-06
Add more compatibility for unreads

VAWVAW:
I have a problem where my read receipt for the main thread isn't updated if there is already an unthreaded receipt for the same event. I'm not sure if this is the sdk or synapse but it messes up the read receipts in some rooms. This will use either the main thread or the unthreaded receipt depending on which one is newer.

### DIFF
--- a/src/base.rs
+++ b/src/base.rs
@@ -1155,11 +1155,7 @@ impl RoomInfo {
             }
         });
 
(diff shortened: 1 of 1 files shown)


ulyssa:
Makes sense! Thanks for digging into the unreads issues further!

PR #453, opened 2025-06-08
Improve image preview placeholder

VAWVAW:
Show image alt text in the middle of the area to make it more visible when the image is partially off-screen on the top.

Also the text "Loading..." doesn't describe the image not fitting in the visible space for the message and "Cut off..." is a bit more descriptive while not being longer.

### DIFF
--- a/src/message/mod.rs
+++ b/src/message/mod.rs
@@ -176,7 +176,9 @@ fn placeholder_frame(
     }
     let mut placeholder = "\u{230c}".to_string();
     placeholder.push_str(&" ".repeat(width - 2));
-    placeholder.push_str("\u{230d}\n");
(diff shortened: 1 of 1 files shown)


src/message/mod.rs
```diff
@@ -1087,7 +1089,7 @@ impl Message {
                 },
                 ImageStatus::Loaded(backend) => {
                     proto = Some(backend);
-                    placeholder_frame(Some("Loading..."), width, &backend.area().into())
+                    placeholder_frame(Some("Cut off..."), width, &backend.area().into())
```
ulyssa, line 1092: Thank you for fixing this message. I've noticed it before but kept forgetting to do something about it.

PR #425, opened 2025-05-20
Upgrade matrix-sdk to 0.11.0

kenrachynski:
matrix-sdk 0.11 requires rust 1.85

PR #456, opened 2025-06-11
Upgrade matrix sdk 0.12

kenrachynski:
This can supercede #425

While this update compiles clean, there are some runtime issues with threaded messages. They should probably be investigated prior to merging


VAWVAW:
Can you please try to reproduce your issues? I didn't notice any problems.

kenrachynski:
I'm not seeing the same errors today. I was seeing everything in the thread in the room's message list instead of only in the thread.

Weird behaviour is a double read marker for myself, but that's minor.

ulyssa:
Closing now that #485 updated to 0.13.

PR #458, opened 2025-06-14
Open `:members` in the same window by default

VAWVAW:
Afaik `:members` is the only command that opens a split. This is disrupting for a workflow that uses only one window and the jumplist.

The old behavior can be recreated using the macro `:rightb :vs<Enter>:vert :res 30<Enter>:members<Enter>`.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -10,6 +10,7 @@ default_room = "#iamb-users:0x.badd.cafe"
 default_split = "vertical"
 external_edit_file_suffix = ".md"
 log_level = "warn"
+members_split = "vertical"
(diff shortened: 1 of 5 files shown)


ulyssa:
I think I've convinced myself that opening in the same window is the better default. I originally made it a split to allow for a sidebar type of view, but the default width of 30 columns means that it's never quite wind enough for some display names.

Making it open in the same window by default should hopefully be a better flow. I've added a `settings.members_split` option which can force a split instead, and also updated the PR to use `cmd.switch` for generating the action, so that something like `:vert members` can override it.

PR #459, opened 2025-06-15
Reset mode after sending message

VAWVAW:
Reset the vim mode to Normal after sending a message.

relies on ulyssa/modalkit#158


ulyssa:
Thank you for implementing this and ulyssa/modalkit#158! I usually like to stay in `Insert` mode so I can send another message, so I've added a `normal_after_send` config option to help control which behaviour is used.

Issue #128, opened 2023-07-05
Feature request: Themes + Statusline


nerdbude:
Hello. iamb is such a great tool - I love it.
A great feature would be the use of different color schemes.
I want to use my own color scheme in iamb too.

A Vim-like status bar would also be great.
I'm not a Rust programmer, so these are the cautious wishes I have - but I'll continue to use iamb without the features. Thanks very much

ulyssa:
What kind of information are you looking for in the status bar? Something like the ruler to show line/column information, or something else?

nerdbude:
A bit like the Vim Statusline. Stuff like the current Mode (Visual, Insert...), number of total users in a room, maybe the roomname.

ulyssa:
I was thinking today about what defining a theme would look like. I think the relevant fields to expose would be:

- background color
- default text color
- window border type
- window border color w/ focused/non-focused variants
- window title color
- tab title color w/ focused/non-focused variants
- message bar prompt color w/ focused/non-focused variants
- room/spaces/dms list colors w/ read/unread variants
- room/spaces/dms tags colors
- command bar text color

There's probably some other fields I'm missing. There might also be a way to  fold #137 into this (different prompt colors?).

jennydaman:
It would be nice to be able to customize the color of rich text elements. Currently, there is no visual distinction between regular text and `monospace formatted text`. It would be great if the background and foreground color of `monospace formatted text` was configurable. I would set the background color to some dark opaque grey.

PR #460, opened 2025-06-15
Add a gray background to code blocks

### LINKED ISSUE #590: Feat: format inline code different to the rest of the text
Code blocks enclosed in triple ticks are formatted with a border around them, but inline code blocks are formatted the same as the rest of the text.

Element:
Image

iamb:
Image

Would be nice if they also would be distinguishable. Maybe underlined, or also having a border, or slightly different color.

Thank you for the great project. It's really fun to use it!

VAWVAW:
fixes #590

### DIFF
--- a/src/message/html.rs
+++ b/src/message/html.rs
@@ -370,6 +370,7 @@ impl StyleTreeNode {
                 }
             },
             StyleTreeNode::Code(child, _) => {
+                let style = style.bg(Color::Indexed(236));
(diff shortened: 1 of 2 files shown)


VAWVAW:
The background color only works with white text. We might be able to detect the terminal color or make it configurable via #128.

VAWVAW:
closing in favor of #472

PR #463, opened 2025-06-16
Fix display of tabs in code blocks

VAWVAW:
see `#iamb-users` for context

This converts tabs to spaces according to the new config option `tabstop`.


src/message/html.rs
```diff
@@ -1415,13 +1415,14 @@ pub mod tests {
         let s = concat!(
             "<pre><code class=\"language-rust\">",
             "fn hello() -&gt; usize {\n",
+            " \t// weired\n",
```
ulyssa, line 1418: This is a great test, thank you! :pray:

PR #464, opened 2025-06-17
Render image previews lazily

VAWVAW:
Profiling the startup of iamb in `foot` using `flamegraph`, I noticed that most of the cpu time was spent rendering image previews. This PR changes the rendering to be in the background (not blocking ui responsiveness) and lazily (on message rendering).

~I had some bugs if the last message is an image:~
- ~Scrolling with ^E and ^Y doesn't work if the last message is focused~
- ~if only one image is loaded and the rest of the loaded events are reactions, older events don't load~

# Profiling data
before optimizations:
flamegraph-original
with lazily computed sixel preview:
flamegraph-lazy-sixel
and lazily decoded jpg:
flamegraph-lazy-jpg

### DIFF
--- a/docs/iamb.5
+++ b/docs/iamb.5
@@ -138,6 +138,13 @@ Enable image previews and configure it.
 An empty object will enable the feature with default settings, omitting it will disable the feature.
 The available fields in this object are:
 .Bl -tag -width Ds
+.It Sy lazy_load
+If
+.Sy true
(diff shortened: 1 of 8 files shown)


benjajaja (approved):
Thank you!

Just tested it, works smooth as butter.

Image

ulyssa (approved):

PR #466, opened 2025-06-18
Make blockquotes in message visually distict

VAWVAW:
Before, quotes are just rendered as an indentation. This adds a gray vertical bar

> like this.

The PR doesn't change the text color because it is not needed to recognize a quote imo.


ulyssa:
This looks much better, thank you!

PR #471, opened 2025-06-25
Make merging of options consistent

VAWVAW:
see `#iamb-dev`


krachynski:
ulyssa:
>vaw:
>In Tunables::merge in every attribute the left side takes precedence but for sort and users the right side takes precedence. Is this intended?
No, merge_sorts is doing it backwareds. a should probably be named profile and b renamed global when fixing it to make it clearer that it's "use profile if set or use global"

PR #476, opened 2025-07-08
Recognise URLs in plain text message bodies

### LINKED ISSUE #370: Recognition of URLs is unreliable
I have this situation:
screenshot-2024-11-06_15-11-1730901807

Talking about the left window (with [#python:opensuse.org](matrix:/r/python:opensuse.org)) and the bugzilla links are broken into so many lines, that my terminal doesn’t know what to do with them, and they don’t have any numbers attached to them either. The link in the other window (to https://ffmpeg-api.com/learn/ffmpeg/recipe/picture-in-picture seems) to be recognized just fine.

VAWVAW:
This PR introduces the `linkify` dependency to gather links if no `"formated_body"` exits. Links are available in `:open` but are not visually changed. It only works for links that include the protocol but I believe this limitation comes from the `url` crate.

fixes #370

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2274,6 +2274,7 @@ dependencies = [
  "image",
  "lazy_static 1.5.0",
  "libc",
+ "linkify",
  "markup5ever_rcdom",
  "matrix-sdk",
  "mime",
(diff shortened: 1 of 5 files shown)


ulyssa:
Thank you for adding support for this! :tada:

PR #478, opened 2025-07-21
Support `XDG_CONFIG_HOME` on macOS for config directory resolution

delafthi:
This update enhances config directory resolution on macOS by prioritizing the `XDG_CONFIG_HOME` environment variable when it is set (e.g., for nix-home users with `xdg.enable = true`). If `XDG_CONFIG_HOME` is not defined, the fallback remains the default path: `~/Library/Application Support`.

This change improves compatibility with XDG-based configurations on macOS, providing a more consistent experience across different environments.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,6 +2,7 @@
 use std::borrow::Cow;
 use std::collections::hash_map::DefaultHasher;
 use std::collections::{BTreeMap, HashMap};
+use std::env;
 use std::fmt;
 use std::fs::File;
 use std::hash::{Hash, Hasher};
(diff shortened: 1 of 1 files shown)


delafthi:
I don’t think this change will happen, as the maintainers clearly stated in https://github.com/dirs-dev/directories-rs/issues/47 that they won’t cover this use case. Many tools, such as ruff, and others mentioned in [this PR﻿﻿﻿](https://github.com/astral-sh/ruff/pull/11115), handle this by using ﻿`~/.config﻿` entirely. That said, I think making this behavior opt-in on macOS when ﻿XDG_CONFIG_HOME﻿ is explicitly set, is reasonable.

PR #481, opened 2025-07-31
Add config option for playing sound-hints with desktop notifications

Akselmo:
This patch adds a configuration element called `play_sound` (bool, false by default) that when set true, 
it will tell the notification to use sound hint called `message-new-instant` which then plays a sound, 
depending on the platform the app is running in.

(I haven't really used Rust before so bear with me :sweat_smile: )

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -483,6 +483,8 @@ pub struct Notifications {
     pub via: NotifyVia,
     #[serde(default = "default_true")]
     pub show_message: bool,
+    #[serde(default)]
+    pub play_sound: bool,
 }
 
 #[derive(Clone)]
--- a/src/config.rs
+++ b/src/config.rs
@@ -484,7 +484,7 @@ pub struct Notifications {
     #[serde(default = "default_true")]
     pub show_message: bool,
     #[serde(default)]
-    pub sound_hint: String,
+    pub sound_hint: Option<String>,
 }
 
 #[derive(Clone)]
--- a/src/config.rs
+++ b/src/config.rs
@@ -484,7 +484,7 @@ pub struct Notifications {
     #[serde(default = "default_true")]
     pub show_message: bool,
     #[serde(default)]
-    pub play_sound: bool,
+    pub sound_hint: String,
 }
 
 #[derive(Clone)]
--- a/src/notifications.rs
+++ b/src/notifications.rs
@@ -50,7 +50,7 @@ pub async fn register_notifications(
     }
     let notify_via = settings.tunables.notifications.via;
     let show_message = settings.tunables.notifications.show_message;
-    let play_sound = settings.tunables.notifications.play_sound;
+    let sound_hint = settings.tunables.notifications.sound_hint.clone();
     let server_settings = client.notification_settings().await;
     let Some(startup_ts) = MilliSecondsSinceUnixEpoch::from_system_time(SystemTime::now()) else {
         return;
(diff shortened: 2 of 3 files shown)


VAWVAW:
I think it would make more sense to make the config option a string with the sound name. Some people might want a different sound to play and I'm not certain the default sounds on windows and macos use the same name as linux anyways. We could then list the default sound names for each OS in the documentation.

Akselmo:
> I think it would make more sense to make the config option a string with the sound name. Some people might want a different sound to play and I'm not certain the default sounds on windows and macos use the same name as linux anyways. We could then list the default sound names for each OS in the documentation.

Yeah i can do that instead.

Akselmo:
Turned it into a `String`. I don't know if theres some better way to do that though, I have not written much rust.

src/config.rs
```diff
@@ -484,7 +484,7 @@ pub struct Notifications {
     #[serde(default = "default_true")]
     pub show_message: bool,
     #[serde(default)]
-    pub play_sound: bool,
+    pub sound_hint: String,
```
VAWVAW, line 487:
```suggestion
    pub sound_hint: Option<String>,
```
In rust, `None` is usually used to represent that a value is missing since there is a difference between an empty string and no string.

src/notifications.rs
```diff
@@ -115,11 +116,11 @@ async fn send_notification(
     body: Option<&str>,
     room_id: OwnedRoomId,
     store: &AsyncProgramStore,
-    play_sound: bool,
+    sound_hint: String,
```
VAWVAW, line 119:
```suggestion
    sound_hint: Option<&str>,
```
Using only a reference instead of the owned data is only a nitpick since it doesn't make a difference. But in my opinion, functions should only take owned data if they need it since it might reduce cloning if the caller has no owned version.

src/notifications.rs
```diff
@@ -90,7 +91,7 @@ pub async fn register_notifications(
                                     body.as_deref(),
                                     room_id,
                                     &store,
-                                    play_sound,
+                                    sound_hint,
```
VAWVAW, line 94:
```suggestion
                                    sound_hint.as_deref(),
```

src/notifications.rs
```diff
@@ -142,7 +143,7 @@ async fn send_notification_desktop(
     body: Option<&str>,
     room_id: OwnedRoomId,
     _store: &AsyncProgramStore,
-    play_sound: bool,
+    sound_hint: String,
```
VAWVAW, line 146:
```suggestion
    sound_hint: Option<&str>,
```

src/notifications.rs
```diff
@@ -151,8 +152,8 @@ async fn send_notification_desktop(
         .icon(IAMB_XDG_NAME)
         .action("default", "default");
 
-    if play_sound {
-        desktop_notification.sound_name("message-new-instant");
+    if !sound_hint.is_empty() {
```
VAWVAW, line 155:
```suggestion
    if let Some(sound_hint) = sound_hint {
```

src/notifications.rs
```diff
@@ -151,8 +152,8 @@ async fn send_notification_desktop(
...
+        desktop_notification.sound_name(sound_hint.as_str());
```
VAWVAW, line 156:
```suggestion
        desktop_notification.sound_name(sound_hint);
```

src/tests.rs
```diff
@@ -196,7 +196,7 @@ pub fn mock_tunables() -> TunableValues {
             enabled: false,
             via: NotifyVia::default(),
             show_message: true,
-            play_sound: false,
+            sound_hint: String::from(""),
```
VAWVAW, line 199:
```suggestion
            sound_hint: None,
```

VAWVAW (changes requested):
Looks good. Just some small suggestions according coding style.

Akselmo:
Will write documentation about it next.

VAWVAW (approved):

VAWVAW:
@ulyssa will you also add it to the man page or should that be a new PR?

PR #494, opened 2025-08-22
Handle attachment file names more robustly

### LINKED ISSUE #491: Image preview doesn't work if the image name is an invalid file name
An image with a `/` in its name has no preview and `:open` results in `Input/Output Error: No such file or directory (os error 2)` on linux.

VAWVAW:
fixes #491


ulyssa (approved):

ulyssa:
Good catch! Thanks for fixing this! :sparkles:

PR #508, opened 2025-09-11
Open Matrix URIs on the command line and support user mentions

### LINKED ISSUE #249: tab completion of user names
When in insert mode, it would be extremely convenient if pressing the Tab key autocompleted the current word as the name of some member of the room you're in.

I find this generally useful, but for me it raises to life saver in some contexts: when taking live minutes of a meeting in a chat room, typing people's name can be very time consuming without this, or trivial with it.

Some alternative key combination may be alright, but I already have so much muscle memory that it'd be really nice if it was Tab.

People actually wanting to want to type tabulations into a chat client is exceedingly rare (many clients don't even support it), so we wouldn't really lose anything.

### LINKED ISSUE #393: [Feature] Format mentions in outgoing messages
From `iamb-users`:

> Well, Element Web crafted this:

```
┌────────────────────────────────────────────────────────────────────┐
│ "content": {                                                       │
│    "body": "so if I try Full Name , what does the event            │
│look like?",                                                        │
│    "format": "org.matrix.custom.html",                             │
│    "formatted_body": "so if I try <a href=\"https://matrix.to/#/@  │
│user:server.net\">Full Name</a> , what does the                     │
│event look like?",                                                  │
│    "m.mentions": {                                                 │
│      "user_ids": [                                                 │
│        "@user:server.net"                                          │
│      ]                                                             │
│    },                                                              │
│    "msgtype": "m.text"                                             │
│  }                                                                 │
└────────────────────────────────────────────────────────────────────┘
```

> While iamb crafted this:

```
┌────────────────────────────────────────────────────────────────────┐
│"content": {                                                        │
│    "body": "e.g. @user:server.net (you should see this message as  │
│not displayed correctly on clients other than iamb)",               │
│    "format": "org.matrix.custom.html",                             │
│    "formatted_body": "<p>e.g. @user:server.net (you should see this│
│ message as not displayed correctly on clients other than iamb)</p>\│
│n",                                                                 │
│    "msgtype": "m.text"                                             │
│  }                                                                 │
└────────────────────────────────────────────────────────────────────┘ 
```

> two major differences here: the formatted body had an
> actual link in it and EW added m.mentions which iamb
> does not do.
> feature request, I guess
> interesting that EX did the extra work of spotting the
>  mention and wrapping it in a pill

Given the lack of consistent support among clients, sticking to the centralized, domain-dependent `https://matrix.to` was recommended over the decentralized, domin-independent URI schema `matrix:`.

### LINKED ISSUE #418: Opening a link to a matrix thread should open it within iamb.
Currently, if someone uses a link to share a thread using the "share" link (`https://matrix.to/#/!room-id:so-and-so.com/$message_id?via=so-and-so.com`, e.g.), iamb opens up a browser to the link instead of just opening the room. AFAIK other clients generally parse out matrix share links and navigate to the room/thread in question, at least Element on mobile does.

### LINKED ISSUE #506: Handle Matrix URIs from CLI
Other software (e.g., a browser, a PDF reader, or an email client) might try to open a Matrix URI with iamb. I believe iamb's CLI interface on startup should add an optional URI argument specifying an action to be run on startup, according to https://spec.matrix.org/latest/appendices/#uris. Note that this includes both `matrix:` and `https://matrix.to` URIs.

VAWVAW:
This changes completion of matrix ids to be a matrix.to link. I don't think there is a better solution because we need a link for mentions to work.

To mention a user complete either an empty word or an `@` and the start of the username or displayname. This will complete a markdown formatted `matrix.to` link for the user. When sending the message, all `matrix.to` links are added as intentional mentions. It is not possible to mention a room.

I'm not sure where to add the documentation for mentioning a user. `iamb.1`? [iamb.chat](https://iamb.chat)?

fixes #506
fixes #418
fixes #393
fixes #249

### DIFF
--- a/docs/iamb.1
+++ b/docs/iamb.1
@@ -5,7 +5,7 @@
 .\"
 .\" You can preview this file with:
 .\"     $ man ./docs/iamb.1
-.Dd Mar 24, 2024
+.Dd Sep 11, 2025
 .Dt IAMB 1
 .Os
 .Sh NAME
@@ -16,6 +16,7 @@
 .Op Fl hV
 .Op Fl P Ar profile
 .Op Fl C Ar dir
+.Op Ar URI
 .Sh DESCRIPTION
 .Nm
 is a client for the Matrix communication protocol.
@@ -46,6 +47,9 @@ Show the help text and quit.
 Show the current
 .Nm
 version and quit.
+.It Ar URI
+A matrix uri or matrix.to link to open on startup. Specified at
+.Lk https://spec.matrix.org/latest/appendices/#uris
 .El
 
 .Sh "GENERAL COMMANDS"
--- a/iamb.desktop
+++ b/iamb.desktop
@@ -1,7 +1,8 @@
 [Desktop Entry]
 Categories=Network;InstantMessaging;Chat;
 Comment=A Matrix client for Vim addicts
-Exec=iamb
+Exec=iamb %u
+MimeType=x-scheme-handler/matrix
 GenericName=Matrix Client
 Keywords=Matrix;matrix.org;chat;communications;talk;
 Name=iamb
--- a/src/base.rs
+++ b/src/base.rs
@@ -1024,12 +1024,34 @@ impl RoomInfo {
 
     /// Get an event for an identifier.
     pub fn get_event(&self, event_id: &EventId) -> Option<&Message> {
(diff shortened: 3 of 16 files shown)


lunasophia:
I've been using this PR for a bit and it works as expected. Would love to see it merged if possible :)

PR #510, opened 2025-09-12
Implement basic reloading

### LINKED ISSUE #505: [Feature Request] Hot Reload Config
Something akin to how Swaywm and Alacritty behave. The user can set an option in the config file and run a command to reload iamb with the new settings without quitting and relaunching. Perhaps a `:source` or `:source-config` command.

If there's some error in the config, iamb provides `:messages` as to what those are and falls back to the previous config state.

Additonally, `:edit-config` or similar command could be used to invoke `$EDITOR` and persist the current iamb session. When the file is saved and closed, iamb reloads with the new config, barring any errors the user made.

VAWVAW:
This adds `:reload` to reload some config options.

Options not currently affected:
- `log_level`
- image previews
- notifications
- ...

Further testing is needed to see which options need to be handled in code.

fixes #505

### DIFF
--- a/docs/iamb.1
+++ b/docs/iamb.1
@@ -57,6 +57,8 @@ View a list of direct messages.
 .It Sy ":logout [user id]"
 Log out of
 .Nm .
+.It Sy ":reload"
+Reload the config.
 .It Sy ":rooms"
 View a list of joined rooms.
 .It Sy ":spaces"
--- a/src/base.rs
+++ b/src/base.rs
@@ -539,6 +539,9 @@ pub enum IambAction {
 
     /// Clear all unread messages.
     ClearUnreads,
+
+    /// Reload the config.
+    ReloadConfig,
 }
 
 impl IambAction {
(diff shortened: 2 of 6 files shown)


fuguesoft:
I'll have a look at this

fuguesoft:
@VAWVAW 
When building with rust, where do I need to configure things?
It doesn't seem to be using the default location but I'm new to this so it's probably user error

fuguesoft:
Okay yeah maybe I built this wrong.
There doesn't seem to be a reload command it can find.

I pulled your fork down and checked out the `hot-reload` branch.
Then I did:

```sh
cargo install --locked iamb
```

fuguesoft:
I got it and from what I can see, the notifications don't seem to be resetting
keymaps seem to change on reload

Are there ways to send you logs or some sort of debug for this?

fuguesoft:
ah okay I'll make another profile and do some more testing

so far I don't see notification settings updating

VAWVAW:
closing in favor of #578

PR #513, opened 2025-09-18
Document how to build from sources

strk:
It wasn't clear to me (not a Rust developer) but also wasn't clear to people who gave me instructions in the matrix room for users... so, I think it's good to document it in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,16 +51,19 @@ url = "https://example.com"
 user_id = "@user:example.com"
 ```
 
-## Installation (building)
+## Installation (from source)
 
-Install Rust (1.83.0 or above) and Cargo, and then run:
+Install Rust and Cargo using [rustup], and then run from the directory
+containing the sources (ie: from a git clone):
 
-From the directory containing the sources (ie: from a git clone):
 ```
 cargo install --locked --path .
 ```
 
-Via `crates.io` (installs the latest tagged release on `crates.io`):
+## Installation (via `crates.io`)
+
+Install Rust (1.83.0 or above) and Cargo, and then run:
+
 ```
 cargo install --locked iamb
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,10 +51,16 @@ url = "https://example.com"
 user_id = "@user:example.com"
 ```
 
-## Installation (via `crates.io`)
+## Installation (building)
 
 Install Rust (1.83.0 or above) and Cargo, and then run:
 
+From the directory containing the sources (ie: from a git clone):
```
lnicola, line 58: I'd move this to another section ("Installation (from local checkout)" or something like that). `cargo install` here just builds and copies the binary to a certain install path.

strk:
Well I've been told, in the Matrix iamb-users room: `weirdrock | there hasn't been a release in about a year now ` and suggested to `weirdrock | try installing from git`, which is why I've been looking for information about doing that...

lnicola:
Ah, but that's `cargo install --git https://github.com/ulyssa/iamb`. The only downside is longer compile times because it's not going to cache the build artifacts.

strk:
If I have a local checkout (as the documentation says) I'm planning to build again on next pull so artifacts I want them...
Maybe you want to add another line, which I may agree on, to build from lastest source using the `--git https://github.com/ulyssa/iamb` line but I'll wait for @ulyssa to tell to avoid wasting time

ulyssa:
Thank you for adding this! 

> I didn't want to repeat the Rust version requirement and the configuration reference, thus the merge of the two, which have "build from source" in common, right?

As long as someone is using `cargo` installed via `rustup`, then it should download the right Rust version by reading the contents of `rust-toolchain.toml`. I've made an edit to mention `rustup`, and moved it into its own sectoin

PR #516, opened 2025-09-19
Hide incorrectly displayed old messages in scrollback

VAWVAW:
Since state events were added, the top of the scrollback is cluttered with room create events making scrolling to the top to load older messages junky because they are inserted bellow the cursor.

This PR hides all messages that were not fetched in `load_older_one` making the scrollback behave more predictable.


VAWVAW:
This is a naive implementation that breaks in some edge cases.

PR #517, opened 2025-09-19
Add MSC4357 live messaging support

DonPrus:
Implements [MSC4357](https://github.com/matrix-org/matrix-spec-proposals/pull/4357) live messaging protocol for real-time message composition in iamb. Messages appear gray while being typed and update in real-time for all room participants.

## Features

- `/live <message>` command to compose messages that update as you type
- Live messages appear gray (RGB 100,100,100) to indicate they're still being composed
- Real-time updates visible to all room participants
- Messages finalize to normal color when Enter is pressed
- Rate limiting to prevent excessive server load

## Implementation Details
- **Live message detection**: Uses `org.matrix.msc4357.live` marker in message JSON
- **Raw event handling**: Preserves custom JSON fields that would be lost during deserialization
- **Session management**: `LiveMessageManager` tracks active live sessions per room
- **Proper edit handling**: Only the latest edit determines if a message is live (prevents old messages from appearing gray after restart)

## Bug Fixes

- Fixed issue where all historical edits were processed instead of just the latest one, which caused old messages to incorrectly appear as live after client restart

## Changes

- Added `src/message/live.rs` - Core live messaging functionality
- Added `src/message/live_detection.rs` - Centralized marker detection
- Modified worker to handle raw message events and track live status
- Updated chat window to handle `/live` command and send updates
- Messages with live marker render in gray color

https://github.com/user-attachments/assets/e70aee7b-0e9d-479d-acd4-c0583c5ae2ac

### DIFF
--- a/src/base.rs
+++ b/src/base.rs
@@ -178,6 +178,7 @@ pub enum MessageAction {
     /// and error when it doesn't recognize it. The second [bool] argument forces it to be
     /// interpreted literally when it is `true`.
     Unreact(Option<String>, bool),
+
 }
 
 /// An action taken in the currently selected space.
@@ -898,6 +899,9 @@ pub struct RoomInfo {
     /// A map of message identifiers to thread replies.
     threads: HashMap<OwnedEventId, Messages>,
 
+    /// Set of event IDs for messages that are live (being typed)
+    pub live_message_ids: HashSet<OwnedEventId>,
+
     /// Whether the scrollback for this room is currently being fetched.
     pub fetching: bool,
 
@@ -929,6 +933,7 @@ impl Default for RoomInfo {
             user_receipts: Default::default(),
             reactions: Default::default(),
             threads: Default::default(),
+            live_message_ids: Default::default(),
             fetching: Default::default(),
             fetch_id: Default::default(),
             fetch_last: Default::default(),
@@ -1087,6 +1092,7 @@ impl RoomInfo {
         let event_id = msg.event_id;
         let new_msgtype = msg.new_content;
 
+
         let Some(EventLocation::Message(thread, key)) = self.keys.get(&event_id) else {
             return;
         };
@@ -1103,9 +1109,12 @@ impl RoomInfo {
             return;
         };
 
+        // Update live status based on live_message_ids
+        msg.is_live = self.live_message_ids.contains(&event_id);
+
         match &mut msg.event {
             MessageEvent::Original(orig) => {
-                orig.content.apply_replacement(new_msgtype);
+                orig.content.apply_replacement(new_msgtype.clone());
             },
             MessageEvent::Local(_, content) => {
                 content.apply_replacement(new_msgtype);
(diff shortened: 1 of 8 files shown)


DonPrus:
I'll double check this at weekends

VAWVAW:
The notification handler should be modified to exclude live messages.

And can't most of the new logic in `worker.rs` in `load_insert` be replaced by looking at the bundled relations of the message?

DonPrus:
> There is `RoomMessageEvent::Original.unsigned.relations.replace`. We could extract that from the raw message in `load_older_one`. But I'm not sure whether the server sends these bundled messages for encrypted rooms.
> 
> If we can use this, we could filter edits events from the message load to preserve bandwidth and increase load times. With live messages there would be more edit events in the history and only the newest one is important anyway.

Thanks for pointing me in the right direction! I investigated this further and you're absolutely right - the server does send bundled relations in unsigned.m.relations.m.replace. I added some logging to check what's actually coming through.

The good news is that for unencrypted rooms, the bundled relations contain the full content of the latest edit, which we could definitely use to skip loading intermediate edits. This would be especially helpful for live messages since they may generate so many edits.

The bad news is that for encrypted rooms, the bundled relations only contain the encrypted ciphertext. Makes sense when you think about it - the server can't aggregate content it can't decrypt, so this optimization wouldn't work for E2E encrypted rooms.

There's also a minor SDK limitation - after deserializing to AnyTimelineEvent, we lose direct access to the unsigned.relations field. We'd have to parse the raw JSON like we're already doing for live marker detection.

If you think it's worth it, I could implement this optimization for unencrypted rooms. I'll check for bundled relations in the raw event, track which events already have the latest content bundled, and skip loading the intermediate edits that are superseded. For original messages with bundled edits, we'd apply that content directly instead of waiting for the separate edit events.

VAWVAW:
> The bad news is that for encrypted rooms, the bundled relations only contain the encrypted ciphertext. Makes sense when you think about it - the server can't aggregate content it can't decrypt, so this optimization wouldn't work for E2E encrypted rooms.

Could we use [`Room::decrypt_event`](https://docs.rs/matrix-sdk/latest/matrix_sdk/room/struct.Room.html#method.decrypt_event) for that?

> If you think it's worth it, I could implement this optimization for unencrypted rooms. I'll check for bundled relations in the raw event, track which events already have the latest content bundled, and skip loading the intermediate edits that are superseded. For original messages with bundled edits, we'd apply that content directly instead of waiting for the separate edit events.

It is your PR and I'm not a maintainer so that is your decision.
But I think that this MSC will generate a lot more edit events and we could probably improve the load time if we manage to filter edit events in [`Room::messages`](https://docs.rs/matrix-sdk/latest/matrix_sdk/room/struct.Room.html#method.messages).

DonPrus:
> It is your PR and I'm not a maintainer so that is your decision. But I think that this MSC will generate a lot more edit events and we could probably improve the load time if we manage to filter edit events in [`Room::messages`](https://docs.rs/matrix-sdk/latest/matrix_sdk/room/struct.Room.html#method.messages).

Since messages come in reverse chronological order, the first edit we encounter is the latest. This commit:
- Tracks which edits we've already seen the latest version
- Skips loading older edits when we already have a newer one
- Skips edits for messages that have bundled replacements
- Significantly reduces the number of events to process, especially for live messages

This optimization works for both encrypted and unencrypted rooms, as we're just filtering duplicate edits regardless of content.

PR #522, opened 2025-10-01
Indicate encryption state of room in messagebar

### LINKED ISSUE #137: Indicate when a message will be sent encrypted
Element will place a green lock in the message bar if it will be sent encrypted, which can be helpful when you're unsure whether a room is encrypted, or whether your client will actually do encryption.

VAWVAW:
This replaces the prompt of the message bar with a green `🔒︎` for encrypted rooms and a red `🔓︎` for unencrypted rooms.

fixes #137


ulyssa (approved):
This looks great! Thank you for implementing this!

PR #529, opened 2025-10-09
Remove blocking timeout for first sync on startup

VAWVAW:
Remove the 30 second timeout if the homeserver had no updates since the last run.

This significantly improves the load times when rapidly restarting the client


ulyssa:
Makes sense, especially for users who aren't in any particularly active rooms.

PR #536, opened 2025-10-13
Add short command aliases

### LINKED ISSUE #535: Feature request: short command names
In the spirit of vim where `:h` can mean `:help`, it would be nice to have some shorthands for frequent commands, e.g. `:c` for `:chats`, `:r` for `:rooms`, `:u` for `:upload`

VAWVAW:
fixes #535

### DIFF
--- a/docs/iamb.1
+++ b/docs/iamb.1
@@ -50,20 +50,20 @@ version and quit.
 
 .Sh "GENERAL COMMANDS"
 .Bl -tag -width Ds
-.It Sy ":chats"
+.It Sy ":c[hats]"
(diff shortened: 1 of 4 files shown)


ulyssa:
Unfortunately, a lot of these are short aliases that could reasonably be added in the future, and that people might want to use because of how common they are in Vim, like `:s` to make replacements on the current line, or `:r` to read in an external file into the message bar.

I think the better approach is probably to allow people to specify their own aliases in their configuration file. I've made https://github.com/ulyssa/modalkit/pull/220 so that that can be done here in `iamb`.

PR #538, opened 2025-10-14
Expand `~` and shell variables in dirs config

### LINKED ISSUE #244: Expand `~` in directory overrides in configuration
@foxyseta noticed that the `dirs` values don't expand a leading `~/` to the home directory. It would be nice to do so to make configurations more portable and to also avoid surprises.

VAWVAW:
fixes #244


ulyssa (approved):

ulyssa:
Thank you for implementing this! :pray:

PR #545, opened 2025-10-26
Fix CI on main branch

VAWVAW:

PR #547, opened 2025-10-26
Bump MSRV to 1.89

ulyssa:
There seems to be a weird hash conflict with the Nix workflow since bumping `rust-toolchain.toml` to be `1.88`, so I'm going to try `1.89` to see if that fixes the issue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ license = "Apache-2.0"
 exclude = [".github", "CONTRIBUTING.md"]
 keywords = ["matrix", "chat", "tui", "vim"]
 categories = ["command-line-utilities"]
-rust-version = "1.88"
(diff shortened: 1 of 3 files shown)


VAWVAW:
The hash of the file needs to be updated. (see #545 )

benjajaja:
Normally the hash only should need updating if rust-toolchain.toml changes, and the new hash is what nix says it has gotten ([example](https://github.com/ulyssa/iamb/actions/runs/18819471428/job/53692701873#step:5:485)). It goes [here](https://github.com/ulyssa/iamb/blob/main/flake.nix#L33).

But that's only if you want to fix it yourself in the same PR where rust-toolchain.toml changes, otherwise some nix people will surely fix it soon enough :smile:

Issue #557, opened 2025-12-21
Respect XDG_CONFIG_DIR on macOS


dancojocaru2000:
If XDG_CONFIG_DIR is set, it should be used instead of Application Support on macOS

PR #568, opened 2026-02-01
Remove redundant check of XDG_CONFIG_HOME

couldbeanemail:
When checking the config iamb checks XDG_CONFIG_HOME twice. 
Once with the dirs dependency. Once on its own. This removes the redundant checks, and the now unused function.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -875,14 +875,9 @@ pub struct ApplicationSettings {
 }
 
 impl ApplicationSettings {
-    fn get_xdg_config_home() -> Option<PathBuf> {
-        env::var("XDG_CONFIG_HOME").ok().map(PathBuf::from)
-    }
-
     pub fn load(cli: Iamb) -> Result<Self, Box<dyn std::error::Error>> {
         let mut config_dir = cli
             .config_directory
-            .or_else(Self::get_xdg_config_home)
             .or_else(dirs::config_dir)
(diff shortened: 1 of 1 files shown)


VAWVAW:
`dirs` only checks for `XDG_CONFIG_HOME` on linux. The current version allows to override the config dir with environment variables on all platforms. I'm not sure whether that is intended though.

couldbeanemail:
Afaik MacOS doesn't really support XDG_CONFIG_HOME. Thus it wouldn't
really be well defined to support it anyways. There seem to be other
mechanisms.

See:
https://github.com/dirs-dev/dirs-rs/issues/56
https://github.com/dirs-dev/directories-rs/issues/47#issuecomment-478337412

In the case that one would want to pretend that XDG variables work 
on other platforms, one should remove the call to dirs and make
this explicit.

best regards :)

couldbeanemail:
Also: the decision should be in line with what is decided in #557 or
vice versa.
I think the same logic would apply there, but I am aware that people
have different takes on this.

ulyssa:
This exists because it is needed for folks using nix `home-manager` on macOS, where [xdg.enable](https://nix-community.github.io/home-manager/options/home-manager/xdg.html#opt-xdg.enable) can be set to allow managing everything at the XDG config paths, which can then have its files managed by `home-manager`.

Issue #80, opened 2023-04-09
Add support for routing through SOCKS proxies


ghost:
Is there a recommended way for routing iamb client's connection through tor socks5 proxy?

VAWVAW:
According to https://docs.rs/reqwest/latest/reqwest/#proxies, we could just enable the `socks` feature of reqwest. The only question is whether we should add a config option or just rely on the users setting `all_proxy`.

KaraZajac:
Hello! I have added a PR to help fix this issue (https://github.com/ulyssa/iamb/pull/575). I wanted to use iamb, but I also needed to use a proxy. The fix works on my macOS machine and lets me connect to my Matrix server (which runs on a Tor .onion service). I hope it helps! I did have AI help me write it. But I manually updated the code to conform to the project's overall style. 

Thank you. :)

PR #575, opened 2026-02-23
Add SOCKS5 proxy support via `[proxy]` config

KaraZajac:
### Summary

Adds optional proxy support for Matrix HTTP traffic using reqwest’s `socks` feature and a new `[proxy]` section in the config. Users can set a global proxy or override it per profile. If no proxy is configured, behavior is unchanged (no proxy and no use of `ALL_PROXY`).

### Changes

**Cargo**
- **`Cargo.toml`**: Depend on `reqwest` with the `socks` feature so the client can use SOCKS5 (and other proxy URLs reqwest supports).

**Configuration**
- **`src/config.rs`**:
  - New `ProxyConfig` with optional `url` and a `merge()` for profile override.
  - `IambConfig` and `ProfileConfig` each have an optional `proxy: Option<ProxyConfig>`.
  - `ApplicationSettings` gains `proxy_url: Option<String>`; in `load()`, global and profile proxy are merged (profile overrides global) and the result is stored in `proxy_url`.

**HTTP client**
- **`src/worker.rs`**: In `create_client_inner`, when `settings.proxy_url` is set, the reqwest client is built with `reqwest::Proxy::all(proxy_url)`. Invalid proxy URL is reported with a clear `.expect(...)` message at startup.

**Docs and examples**
- **`config.example.toml`**: Added a commented `[proxy]` example with `url = "socks5://127.0.0.1:1080"`.
- **`docs/iamb.5`**: New **PROXY** section documenting `[proxy]` and `url`, and `proxy` listed as overridable per profile.

**Tests**
- **`src/tests.rs`**: `mock_settings()` updated so `ProfileConfig` includes `proxy: None` and `ApplicationSettings` includes `proxy_url: None`.

### Usage

- **Global**: add `[proxy]` with `url = "socks5://host:port"` (or another supported proxy URL).
- **Per-profile**: set `proxy = { url = "socks5://..." }` under a profile; it overrides the global proxy for that profile.
- Omit `[proxy]` or leave `url` unset to use no proxy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1540,6 +1540,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "encoding_rs"
+version = "0.8.35"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "75030f3c4f45dafd7586dd6780965a8c7e8e285a5ecb86713e63a79c5b2766f3"
+dependencies = [
+ "cfg-if",
+]
+
+[[package]]
 name = "endian-type"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2389,9 +2398,11 @@ dependencies = [
  "percent-encoding",
  "pin-project-lite",
  "socket2 0.6.1",
+ "system-configuration",
  "tokio",
  "tower-service",
  "tracing",
+ "windows-registry",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


VAWVAW:
would fix #80 if this wasn't AI slop that ignores the coding style of this project and doesn't even compile (without openssl available).

KaraZajac:
Hey thanks! I updated the styling and removed the OpenSSL dependency! I hope it looks better. You are correct that I did have AI's help in creating this.  Thank you for taking the time to review the code.

ulyssa:
I've used this and #576 to put together #640, which should have all of the same features as both, although it uses a `[settings.proxy]` subsection instead of `[proxy]`.

PR #576, opened 2026-02-25
Socks proxy support

### LINKED ISSUE #80: Add support for routing through SOCKS proxies

VAWVAW:
Support the use of `socks5://` proxies and document the usage of proxies.

Also add `settings.proxy` to specify a proxy per profile or ignore the system proxy (set to empty string).

fixes #80
alternative to #575

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2432,6 +2432,7 @@ dependencies = [
  "ratatui",
  "ratatui-image",
  "regex",
+ "reqwest",
  "rpassword",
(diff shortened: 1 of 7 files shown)


ulyssa:
I've used this and #575 to put together #640, which should have all of the same features as both, although it uses a `[settings.proxy]` subsection instead of being just a string with the URL.

PR #520, opened 2025-09-21
Improve completions

VAWVAW:
Add helpful completions for all iamb commands and improve `:space child set` to accept a room alias like `:join`. ~Also add missing documentation for `:room history`~

~relies on ulyssa/modalkit#188 (and ulyssa/modalkit#186 for file paths)~ (merged)

PR #578, opened 2026-02-25
Add commands `:set` and `:reload`

### LINKED ISSUE #177: add option to disable case sensitivity when searching rooms
In my vimrc, I `set ic` (ignorecase) because I'm lazy and can't be bothered to remember the exact casing used for whatever I'm trying to find. Currently when searching a list of rooms (and probably other stuff too?), the search is case-sensitive.  An option to disable this would be really great!

### LINKED ISSUE #367: request: ability to configure settings from within the client
i'm just playing with iamb for the first time today (fun!), but it's a little obnoxious to mess with the app, realize i want to change a setting, edit the config in another terminal, quit iamb, start it again, wait for it to sync, get hit with [this bug](https://github.com/ulyssa/iamb/issues/292) and have to check another client to make sure i'm not missing anything important before running `:unreads clear`, and then do it all over again about three minutes later.

most tui apps like iamb offer the ability to set the configuration dynamically, perhaps with a `:set settings.image_preview.protocol.type=kitty` command structure, e.g.

if this were vim, that setting would of course only work for the current session... but if it were a more intelligent tui app, it could also store this configuration automatically into the already-loaded config.toml, to be reloaded next time the app was opened.

again, if it were vim, at the very least there would be a command to reload the config file. this is less ideal, but may be simpler to implement.

### LINKED ISSUE #505: [Feature Request] Hot Reload Config

### LINKED ISSUE #567: Notification toggle command
Desktop notifications can currently be configured in the config, but it does not seem possible to then disable notifications temporarily in a running iamb instance without modifying the config. It would be useful if we had a command such as `:notifications {on,off,toggle}` that could be used to temporarily turn on and off notifications when they're already configured in the config file.

VAWVAW:
Allow the config to be reloaded and specific settings to be changed at runtime.

Most config values under `settings` are changeable and `dirs.downloads` is updated on `:reload`.

todo:
- [ ] `cache_policy`
- [ ] `state_event_display`
  - need to change the logic to store state events that are not displayed
- [ ] `mouse`, `terminal.enable_extended_keys`
  - need to modify the panic hook
- [x] improve completions and argument parsing
  - easier after #520
- [ ] add tests

Unsupported options:
- `default_room`
  - doesn't make sense to change at runtime
- `image_preview.protocol.font_size`
  - option is deprecated
- `terminal.enable_title`
  - I think there is no way to restore old value
- `max_log_files`, `request_timeout`, `ssl_verify`, `proxy`
  - hard/impossible to change at runtime

Macros should be added in a followup that adds a `:map` command.

fixes #505
supersedes #510
fixes #367
fixes #177 
fixes #567


VAWVAW:
I believe this is useful in its current state and the missing settings can be added in follow up PRs.

PR #579, opened 2026-02-27
Add `:mentions` window and hopefully finally fix unreads

### LINKED ISSUE #376: Sorting does not work correctly
I use both Element and iamb on my computer, and when I use iamb, the sorting isn't the same. It seems to be random, and not based on time or which chat is unread. Is this normal, and if so, can I change this?

### LINKED ISSUE #382: :mentions
The `:unread` feature is very helpful to catch up with rooms, but it can get very crowded when you have joined a lot of rooms.

Either being able to filter out low-priority rooms or having a `:mentions` feature to get the list of rooms where I'm mentioned would be very useful :)

### LINKED ISSUE #553: Messages sent from another client/device mark rooms as unread
When sending a message from another client/device, it makes the room show up in `:unreads` despite it being your message.
This should be fixable by adding to the implementations of `RoomLikeItem::is_unread()` with a check to make sure it's not a message sent from your account that is triggering this status.

[Related issue from the old synapse repository](https://github.com/matrix-org/synapse/issues/14837) (implying that this is something that should indeed be handled by the client)

### LINKED ISSUE #564: All rooms marked as unread
when closing iamb and opening it again, everything that was read becomes unread again, all rooms and dms
im using iamb 0.0.11 and [tuwunel](https://github.com/matrix-construct/tuwunel) as my homeserver (this doesnt seem to happen with synapse)
ref #292

VAWVAW:
Just use sdk functions to track unreads.

And send read receipts according to spec:
- no receipts for our own events
- no receipts for state events
- private read receipts if we don't want public receipts

This also adds the distinction between
- highlights (the user in mentioned)
- notifications (what should be shown as unread)
- messages (in muted rooms)
and doesn't highlight the last one in bold in the room list.

fixes #564
fixes #553

Edit: Also add a `:mentions` window.
fixes #382

### DIFF
--- a/docs/iamb.1
+++ b/docs/iamb.1
@@ -65,6 +65,8 @@ View a list of joined spaces.
 View a list of unread rooms.
 .It Sy ":unreads clear"
 Mark all rooms as read.
+.It Sy ":mentions"
+View a list of rooms with mentions of the current user.
 .It Sy ":welcome"
 View the startup Welcome window.
 .It Sy ":forget"
--- a/docs/iamb.5
+++ b/docs/iamb.5
@@ -212,7 +212,7 @@ Defines whether or not reactions should be shown.
 Defines whether or not reactions should be shown as their respective shortcode.
 
 .It Sy read_receipt_send
-Defines whether or not read confirmations are sent.
+Defines whether or not public read confirmations are sent.
 
 .It Sy read_receipt_display
 Defines whether or not read confirmations are displayed.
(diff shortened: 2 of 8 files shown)


kybe236:
Unreads seem to be cleared when closing and opening iamb with this patch. and sometimes instantly when being in the dms window or switching from unreads to dms or the other way around.

im using tuwunel as homeserver

VAWVAW:
I believe this was an upstream issue and it should be fixed after #625 is merged.

ulyssa:
Thank you for chasing this down! Hopefully just letting the SDK handle this will finally take care of it.

Issue #552, opened 2025-11-29
Text typed into message is placed in the space after the cursor, not the space it's in, but only when images are rendered


Saithe6:
Whenever I try to type into the message box, the text placement is offset one to the right from where the cursor is. If the window is unfocused, the lock icon disappears, and if the window is affected by this bug, the text is shifted one space to the left (to the spot it should be). It seems to only occur when an image is being rendered. Unfortunately I cannot provide screenshots because that unfocuses the window.

I'm running Hyprland 0.45.2 if it matters, and have tested this in xterm and wezterm. I am running the most recent commit compiled with release mode.

Elec3137:
I've seen something like this issue, but even though I don't have image preview enabled which is what I think you're saying is  causing the problem?

Note that you can share screenshots by recording it and then sending the relevant frames from the resulting video

eclairevoyant:
still present on iamb 0.0.11 and `main` (b10c985)

Elec3137:
Is it fixed by https://github.com/ulyssa/iamb/pull/581 ?

eclairevoyant:
yes, it appears so.

eclairevoyant:
fixed by #581, this can be closed.

Issue #551, opened 2025-11-18
Allow customizing the encryption state indicator


baraahomsi:
It would be nice to be able to set custom characters for #522, or at least have a toggle to use ASCII/non-emoji fallback ones.

VAWVAW:
fixed in #684

PR #581, opened 2026-03-03
Support placing encryption indicator in room title instead of prompt

### LINKED ISSUE #569: Strange emoji in message input area instead of `>`
The message input area displays a sort of quarter-circle ahead of the cursor instead of the `>` behind the cursor I see in the demos. Usually the circle is red, but sometimes it's green.

This is with iamb commit 93fc47d, kitty version 0.45.0, continuwuity home server. Locale is `en_US.UTF-8`, font is `Noto Sans Mono`.

Image

VAWVAW:
This is a partial revert of #522

Instead of showing the encryption indicator in the message bar show it in the window title. This should work around some issues where emojis aren't displayed correctly which leads to inconsistent text input.

Also add the config option `encryption_indicator` to disable this feature.

addresses #551
fixes #569 #552

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -21,6 +21,10 @@ typing_notice_send = true
 user_gutter_width = 30
 username_display = "username"
 
+[settings.encryption]
+indicator = "only-encrypted"
+location = "title|prompt"
+
 [settings.image_preview]
 protocol.type = "sixel"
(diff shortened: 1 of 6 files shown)


eclairevoyant:
fixes #552 as well (mind adding that to your PR desc?)

eclairevoyant (approved):

PR #586, opened 2026-03-13
Use `matrix-sdk-ui` for internal representation

### LINKED ISSUE #588: [Feature Request] Add a :invites room
This would be like the `:rooms` list and users can see current invites, then press enter on the invite would bring up a list of actions like:
- Accept or Join
- Reject
- Reject & Block

VAWVAW:
Use [`matrix-sdk-ui`](https://docs.rs/matrix-sdk-ui/latest/matrix_sdk_ui/) for internal statekeeping.
This will reduce inconsistencies with other clients by respecting the spec better (e.g. message order by homerserver order instead of origin server timestamp or which messages should get a read receipt). 

Things this PR provides/does:
- update `modalkit` to newest git
- update `matrix-sdk` from `0.14.0` to `0.16.0`
- make sending nonblocking
- implement `:invites` list
- show replied to message for replies to old messages that are not loaded
- show and update the fully_read marker

Future Possibilities (that are way easier with `matrix-sdk-ui`):
- support polls
- show the raw json for an event (in combination with #566)
- differentiate in ui between unread messages, notifications and highlights (maybe with #472)
- show a [shield state](https://docs.rs/matrix-sdk-common/0.16.0/matrix_sdk_common/deserialized_responses/enum.ShieldState.html) for problematic messages
- going directly to the message context on `:replied` instead of paginating backwards until we hit it

includes #464
provides #445, #516, #579
fixes #588

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,23 +4,14 @@ version = 4
 
 [[package]]
 name = "accessory"
-version = "1.3.1"
+version = "2.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "87537f9ae7cfa78d5b8ebd1a1db25959f5e737126be4d8eb44a5452fc4b63cde"
+checksum = "28e416a3ab45838bac2ab2d81b1088d738d7b2d2c5272a54d39366565a29bd80"
 dependencies = [
  "macroific",
  "proc-macro2",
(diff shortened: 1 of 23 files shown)


VAWVAW:
After spending some time on this, I don't believe this is worth it.

PR #587, opened 2026-03-13
Add configurable cursor shape via `settings.terminal.cursor_shape`

### LINKED ISSUE #197: Feat: Cursor shape options
I'm not used to seeing a full block as my cursor when typing, everywhere it's nowadays a vertical line which splits characters, even in (neo)vim. It would be great to have this as a customisation option in iamb as I see vertical line cursors as the standard for this century.

```
							*tui-cursor-shape*
Nvim will adjust the shape of the cursor from a block to a line when in insert
mode (or as specified by the 'guicursor' option), on terminals that support
it.  It uses the same |terminfo| extensions that were pioneered by tmux for
this: "Ss" and "Se".
Similarly, if you set the cursor highlight group with blend=100, Nvim hides
the cursor through the "cvvis" and "civis" extensions.

If your terminfo definition is missing them, then Nvim will decide whether to
add them to your terminfo definition, by looking at $TERM and other
environment variables.  For the "rxvt", "putty", "linux", "screen",
"teraterm", and "iterm" terminal types, or when Konsole, a libvte-based
terminal emulator, or genuine Xterm are detected, it will add constructed
"Ss" and "Se" capabilities.
```

xsa-dev:
### Summary

Add a new `settings.cursor_shape` option to configure the terminal cursor shape used by iamb.

Supported values:
- `default`
- `block`
- `line`
- `underline`

### Changes

- add `CursorShape` config enum with serde parsing
- add `cursor_shape` to `Tunables` / `TunableValues` (including merge + default handling)
- apply selected shape in TTY setup using `crossterm::cursor::SetCursorStyle`
- restore terminal cursor style to default on exit
- document the new setting in `docs/iamb.5`
- add `cursor_shape = \"default\"` to `config.example.toml`
- add unit test coverage for parsing (`test_parse_cursor_shape`)

### Motivation

Closes a long-standing request to support non-block cursors and allows users to match their terminal/editor preferences.

Closes #197

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -5,6 +5,7 @@ user_id = "@user:matrix.org"
 url = "https://matrix.org"
 
 [settings]
+cursor_shape = "default"
 default_room = "#iamb-users:0x.badd.cafe"
 external_edit_file_suffix = ".md"
 log_level = "warn"
--- a/config.example.toml
+++ b/config.example.toml
@@ -6,7 +6,6 @@ password_file = "/path/to/password"
 url = "https://matrix.org"
 
 [settings]
-cursor_shape = "default"
 default_room = "#iamb-users:0x.badd.cafe"
 external_edit_file_suffix = ".md"
 log_level = "warn"
@@ -35,6 +34,7 @@ rooms = ["favorite", "lowpriority", "unread", "name"]
 members = ["power", "id"]
 
 [settings.terminal]
+cursor_shape = "default"
 enable_extended_keys = false
 enable_title = true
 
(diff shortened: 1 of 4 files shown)


VAWVAW (commented):
This works as described but I'm not sure that this is the best way to implement it. I believe there should at least be an `auto` option to change the cursor shape depending on the vim mode (i.e. `block` in normal mode, `line` in insert mode and `underline` in replace mode). Since I believe that `modalkit` doesn't expose the current mode, it would make more sense to implement this in [`modalkit`](https://github.com/ulyssa/modalkit).

ulyssa:
I've made ulyssa/modalkit#201 so that it will be possible to add `cursor_shape_insert` and `cursor_shape_replace` options once the dependency is updated.

PR #592, opened 2026-03-19
Support media captions

### LINKED ISSUE #606: Upload breaks thread mode
If you do a `:upload /some/file` while inside a thread, the message with that file gets sent outside the thread and all subsequent messages also appear outside the thread. Getting out of the thread (`^o`) and coming back in (`<Enter>`) won't solve it, messages still appear outside.
The only way to be able to send messages inside the thread is to restart iamb.

I am on iamb version `0.0.11`

VAWVAW:
Currently the `body` version of the media caption is rendered as the file name. This PR shows the caption under the file name and supports html captions.

When using `:upload` or pasting an image and the message bar is not empty, the user is queried whether to use the content of the message bar as a caption.

fixes #606

### DIFF
--- a/src/base.rs
+++ b/src/base.rs
@@ -481,10 +481,15 @@ pub enum SendAction {
     SubmitFromEditor,
 
     /// Upload a file.
-    Upload(String),
+    ///
+    /// The second argument indicates whether to use the messagebar as a caption, don't use it or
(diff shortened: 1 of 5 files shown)


VAWVAW:
We should also add support for mentions in the caption after #393 is addressed.

PR #593, opened 2026-03-20
Simplify room creation by using sdk helpers

VAWVAW:


VAWVAW:
These sdk functions are broken in `matrix-sdk 0.14.0` and `matrix 0.16.0`. There is a fix in `main`, so we would need to wait for `0.17.0` to release.

ulyssa (approved):

PR #597, opened 2026-03-23
Add VimBindings::without_nxo_space_motion() for space-prefixed key chords

TheFirstHero6:
### Summary

Adds an opt-in `VimBindings` builder method that removes the default Normal/Visual/Operator-pending binding for a **lone space** (move one column right). That single-key binding prevents applications from registering multi-key sequences whose first key is space (e.g. leader-style chords implemented via config macros). 

### Motivation

Consumers such as [iamb](https://github.com/ulyssa/iamb) map user-defined key sequences after installing default Vim bindings. When space is already a complete `Step`, the input machine never enters a “wait for next key” state for `<Space>` followed by another key, so prefix chords starting with space cannot work.

Removing only this one mapping (there is a single `" "` entry in `default_keys()` today) frees `<Space>` as a chord prefix at the cost of losing the default space-as-motion behavior in NXO modes.

This allows lazyvim style leader navigation in iamb (when coupled with the changes in [modalkit PR](https://github.com/ulyssa/modalkit/pull/193)), as an alternative to the clunkier vanilla vim ":command" format.

### Changes

- `VimBindings::without_nxo_space_motion()` — filters `mappings` to drop the `" "` row.
- Unit test `without_nxo_space_motion_removes_space_mapping`.
- Crate version bump to **0.0.25** (workspace + `modalkit` + `modalkit-ratatui` as applicable).

### API

```rust
let vim = VimBindings::default()
    .submit_on_enter()
    .without_nxo_space_motion();
    
### Breaking Changes
None for default VimBindings::default(); the new behavior is strictly opt-in.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1492,8 +1492,6 @@ dependencies = [
 [[package]]
 name = "editor-types"
 version = "0.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a9e9e99679670f67825fcd24a23cb4eb655a0f92c82bd4d1c1a1357c0cd71e87"
 dependencies = [
  "bitflags 2.10.0",
  "editor-types-macros",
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by VAWVAW** [comment]
What features does this PR provide that aren't possible with a config like this?

```toml
[macros."normal|select|visual"]
"<Space>" = "\\"
"\\e" = ":edit<CR>"
```

With this config pressing `<space>e` edits the selected message.

TheFirstHero6:
Upon reconsideration, ostensibly none past some edge-cases for single-stroke navigation. Bit of an overengineered solution haha, closing PR

PR #601, opened 2026-03-27
Fix `broken-string-escape` warning when using stricter `.nix` parsers 

azizsayfiddin:


ulyssa:
I believe this needs to have the escapes removed instead of added. Without the escapes, the PS1 still renders the same as before:

```
% nix develop
[nix-develop]$ 
```

PR #602, opened 2026-03-27
chore: Update dependencies

VAWVAW:
This bumps the versions for our dependencies including breaking changes.

Notes:
- the documentation for `markup5ever_rcdom` says that it shouldn't be used in production. Maybe we can implement [`TreeSink`](https://docs.rs/html5ever/latest/html5ever/tree_builder/trait.TreeSink.html) directly for our `StyleTree`?
- `ratatui` needs to be updated in sync with `modalkit`
- newer versions for `ratatui-image` include a significant msrv bump and enable a runtime dependency by default

Should I move these notes to a new tracking issue for updates?


ulyssa:
Thank you for updating these!

> the documentation for markup5ever_rcdom says that it shouldn't be used in production. Maybe we can implement [TreeSink](https://docs.rs/html5ever/latest/html5ever/tree_builder/trait.TreeSink.html) directly for our StyleTree?

That might be nice to do. I had picked `markup5ever_rcdom` because it seemed fine for something that wasn't really a web browser. My recollection from when I picked it was that the main issue was "make sure you're not expecting this to be fast", but looking at the docs it seems it's more "we don't guarantee there aren't security issues". 

Of course, that being said, unless the `markup5ever_rcdom` crate explicitly has intentional security holes for testing purposes, it might still be better than maintaining a custom implementation if it ends up being a large amount of code. I'd be curious what your thoughts are on it if you've had a chance to dig into the crate further.

PR #605, opened 2026-03-31
Improve verifications

VAWVAW:
Improve verifications in the following ways:
- refactor state keeping
- don't auto-accept verification requests
  - don't force users to use iamb for verification
- start SAS verification flow
  - verifications between two `iamb` clients were broken because no one would actually `start_sas()`
- show more information about the flow state
  - requests we have sent, requests we can accept, requests waiting for the other user, requests that were completed by another device
- add verification via showing a QR code


ulyssa:
Thank you for cleaning this up, this looks great! :heart: I think most of this hasn't been touched since #1 and it definitely shows sometimes, like with the auto-accept :grimacing:

PR #610, opened 2026-04-21
Add `input_prompt` configuration option for overriding message bar prompt

hanzpo:
when running in tmux, the default 🔒/🔓 encryption indicators didnt work properly for me and displayed a comma then a closing parenthesis. i just added the ability to override the input prompt.

SUMMARY BELOW IS AI GENERATED

## Summary

Adds an optional `input_prompt` string under `[settings]` that overrides the default lock/unlock emoji prompt in the input area.

- **Unset (default):** preserves existing behaviour — 🔒 green for encrypted, 🔓 red for unencrypted, `> ` red for unknown encryption state.
- **Set:** the provided string replaces the prompt for all three states.

## Example

```toml
[settings]
input_prompt = "❯ "
```

## Motivation

When running iamb inside tmux, the default emoji prompt can produce stray characters in the input line that only `Ctrl+L` clears. `tmux capture-pane` on the affected region shows a clean logical buffer — the artifact lives at the terminal display layer, caused by disagreement between ratatui's `unicode-width`, tmux's internal width tables, and the outer terminal's rendering of the lock glyph + `U+FE0E` variation selector. Not configurable at either end.

An ASCII override sidesteps the width mismatch while leaving the emoji prompt intact for users who aren't affected.

## Test plan

- [x] `cargo install --path . --locked` builds clean on rust 1.92
- [x] Default config: emoji prompt + colors render as before
- [x] With `input_prompt = "❯ "` in `[settings]`: single ASCII prompt renders, tmux artifacts disappear
- [x] `config.example.toml` updated with a commented example

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -7,6 +7,7 @@ url = "https://matrix.org"
 [settings]
 default_room = "#iamb-users:0x.badd.cafe"
 external_edit_file_suffix = ".md"
+# input_prompt = "❯ "  # Overrides the input-area prompt. Unset = default 🔒/🔓/> based on encryption state.
 log_level = "warn"
 message_shortcode_display = false
 open_command = ["my-open", "--file"]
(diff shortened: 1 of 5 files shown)


VAWVAW:
Would #581 be a good alternative for this?

If not, please add documentation for the new config option to `docs/iamb.5`.

PR #614, opened 2026-05-04
Accessibility: Specify position of invisible/custom cursor

### LINKED ISSUE #202: accessibility: use real TTY cursor in room list
The `:rooms` list uses a visual cursor instead of the "real" TTY cursor. This
one stays either at the end or the beginning of the list. This is problematic
for blind users using a screen reader. The cursor is used to determine which
item is selected.
You can compare this on GNU/Linux when using dialog, once without
`--visit-items` and once with `--visit-items`. Without this change, a blind user
can just learn the room list and their indices by heart, navigate to the
memorised index and hope they'll end up in the right room.

### LINKED ISSUE #612: Accessibility: improve the use of the system cursor
Assistive technologies like [BRLTTY](https://brltty.app) track the system cursor and will by default display the area of the screen that contains it in braille (or render it through speech).

Therefore, for visually impaired and blind users it is very important that the system cursor is placed in a meaningful way rather than left alone where the last display occurred.

For instance, in the windows that are displayed by the `:chats`, `:rooms` and `:dms` commands, currently the system cursor stays on the bottom-right corner of the window and a blind or visually impaired user has now way to know which line is currently selected. Even worse, without the cursor it was not even clear to me that there actually was a focused line.

I am not sure whether this issue is exhaustive regarding which window do have this issue. Most likely it is not exhaustive, but it would be super helpful ifit could be checked that the system cursor is placed meaningfully in every window.

### LINKED ISSUE #622: Accessibility: improve system cursor placement when navigating messages
This is a follow-up to #612.

Currently, when one moves up through messages with up arrow, the cursor will be places on the line with the day marker rather than being placed on the line where the first message sent that day begins.

This is very surprising at first when reading in braille, also because the day markers are centered on screen so if one uses a large terminal, they will be rater far from its left edge and in particular beyond what a typical 40-cells  braille display can show, so that when the cursor goes on such a line one may be under the impression that the cursor is on an empty line.

VAWVAW:
Not reporting an accurate terminal cursor can break accessibility tools like screen reader or braille displays. Windows like lists that implement a custom highlighting without updating the term cursor can now correctly position the cursor without changing the graphical output.

This change implements the term cursor for `Scrollback` and proxies it to `modalkit-ratatui` types for all other windows.

fixes #612
fixes #202
fixes #622
~relies on ulyssa/modalkit#197~ (merged)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1539,8 +1539,7 @@ dependencies = [
 [[package]]
 name = "editor-types"
 version = "0.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a9e9e99679670f67825fcd24a23cb4eb655a0f92c82bd4d1c1a1357c0cd71e87"
+source = "git+https://github.com/vawvaw/modalkit?rev=c2167c57d2d50e8232ac8ee0e87e9b3454341120#c2167c57d2d50e8232ac8ee0e87e9b3454341120"
 dependencies = [
  "bitflags 2.11.0",
  "editor-types-macros",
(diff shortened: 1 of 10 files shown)


ulyssa:
Thank you for improving this! :pray:

PR #620, opened 2026-06-03
Add `default_split` tunable to control `<C-W>f` split direction

### LINKED ISSUE #415: Open identifier in V Split
Hi, is there a way to open a new window split vertically? I know that `<C-W>f` opens a new window split for the item under the cursor, but is there a way to open it in a vertical split instead? 

some ideas:
- A new shortcut could be `<C-W>g`, perhaps.
- Or, `<C-W>f` could even default to open in vertical splits. I think that vertical columns fit chat information better than many horizontal ones.

stevenpollack:
## Summary

Closes #415.

`<C-W>f` (open item under cursor in a split) currently hardcodes a horizontal split. Users who prefer vertical splits had to work around this with a macro like `"f" = "<C-w>f<C-w><S-l>"`.

This PR adds a `default_split` tunable (`"horizontal"` by default, preserving current behaviour) that, when set to `"vertical"`, overrides `<C-W>f` and `<C-W><C-F>` to open in a vertical split instead.

**Config usage:**
```toml
[settings]
default_split = "vertical"
```

## Implementation notes

- `SplitDirection` enum added to `config.rs`, following the same `#[serde(rename_all = "lowercase")]` pattern as `UserDisplayStyle`.
- `ApplicationSettings::setup()` in `keybindings.rs` calls `bindings.add_mapping` for the two affected key sequences when `default_split = "vertical"`. This works because the underlying `keybindings` crate's `upsert_node` overwrites a node's action when a later registration targets an already-registered key sequence — so it cleanly overrides the horizontal binding that modalkit registers during `VimBindings::setup`.
- No modalkit changes required.

## Test plan

- [x] Build with `cargo build`
- [x] Run `cargo test --locked`
- [x] With `default_split = "vertical"` in config, confirm `<C-W>f` on a room in the room list opens it in a vertical split
- [x] Without the setting (or with `"horizontal"`), confirm existing behaviour is unchanged

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -6,6 +6,7 @@ url = "https://matrix.org"
 
 [settings]
 default_room = "#iamb-users:0x.badd.cafe"
+default_split = "vertical"
 external_edit_file_suffix = ".md"
 log_level = "warn"
 message_shortcode_display = false
--- a/docs/iamb.5
+++ b/docs/iamb.5
@@ -133,6 +133,17 @@ The room to show by default instead of the
 .Sy :welcome
 window.
 
+.It Sy default_split
+Controls which direction
+.Sy <C-W>f
+and
+.Sy <C-W><C-F>
+open a new split when following a room identifier under the cursor.
+Possible values are
+.Dq Sy horizontal
+(the default) and
+.Dq Sy vertical .
+
 .It Sy image_preview
 Enable image previews and configure it.
 An empty object will enable the feature with default settings, omitting it will disable the feature.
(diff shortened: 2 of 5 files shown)


VAWVAW:
Please add documentation to `docs/iamb.5`.

stevenpollack:
@VAWVAW done! Sorry for the omission. Completely slipped my eye.

stevenpollack:
bump for freshness.

VAWVAW:
This is a really slow moving project. It might be a few weeks or months until you get a reaction from a maintainer.

stevenpollack:
> This is a really slow moving project. It might be a few weeks or months until you get a reaction from a maintainer.

fair enough! thanks for the head's up. =)

ulyssa (approved):

PR #627, opened 2026-06-18
Send messages asynchronously using the SDK's send queue

### LINKED ISSUE #199: Slow performance sending in #matrix:matrix.org
To repro:
 * join #matrix:matrix.org
 * send a message
 * on hitting enter; iamb freezes for about 10s
 * eventually the local-italic echo appears, and then resolves to remote echo.

I mistook this earlier as a lack of local echo (or lack of message queuing / retry semantics), but the actual thing which was tripping me up looks to be freeze on hitting enter.

### LINKED ISSUE #359: UI freezes while the client sends messages.
**Description:**
When sending a message, the UI becomes unresponsive. This causes a poor user experience, and it would be better to handle this asynchronously.

**Proposal:**
To improve usability, consider implementing asynchronous processing for message sending.

**Additional Context:**
@ulyssa, I'm willing to work on fixing this issue myself. Could you please provide some guidance on which part of the codebase would be a good starting point?

VAWVAW:
This makes sending messages non-blocking.

### DIFF
--- a/src/base.rs
+++ b/src/base.rs
@@ -13,6 +13,7 @@ use std::time::{Duration, Instant};
 
 use emojis::Emoji;
 use matrix_sdk::ruma::events::receipt::ReceiptThread;
+use matrix_sdk::ruma::OwnedTransactionId;
 use ratatui::{
     buffer::Buffer,
     layout::{Alignment, Rect},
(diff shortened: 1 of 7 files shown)


ulyssa:
This is great work! Thank you for switching the echos over to the SDK's queue!

PR #628, opened 2026-06-28
Add config option for case-insensitive search

codegax:
Adds a `case_insensitive_search` setting (default off) so searching rooms, DMs, spaces, members, and message scrollback ignores case without needing a `(?i)` prefix.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -5,9 +5,9 @@ user_id = "@user:matrix.org"
 url = "https://matrix.org"
 
 [settings]
-case_insensitive_search = false
 default_room = "#iamb-users:0x.badd.cafe"
 external_edit_file_suffix = ".md"
+ignorecase = false
 log_level = "warn"
 message_shortcode_display = false
 open_command = ["my-open", "--file"]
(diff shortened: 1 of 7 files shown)


VAWVAW:
In my opinion it would be more consistent to use the vim naming of `igorecase` for the config option.

And I believe it makes more sense to implement most of this functionality in upstream `modalkit-ratatui` by adding something like `set_ignorecase()` to [`ListState`](https://docs.rs/modalkit-ratatui/latest/modalkit_ratatui/list/struct.ListState.html).

codegax:
Yeah, good points. Renamed it to `ignorecase` so it matches Vim.

I'll put together a PR on modalkit-ratatui to add ignorecase to `ListState`, then switch this over to use it. Only catch is the scrollback search (the `/` one inside a room) builds its regex in iamb rather than modalkit, so I'll probably still need a little something here for that bit.

VAWVAW:
In case you want to keep contributing, I want to warn you that this is a really slow moving project and that it can take a few months until changes are merged.

I'll also point you to the matrix channels [#iamb-users:0x.badd.cafe](https://matrix.to/#/#iamb-users:0x.badd.cafe) and [#iamb-dev:0x.badd.cafe](https://matrix.to/#/#iamb-dev:0x.badd.cafe) and [my fork](https://github.com/vawvaw/iamb) where I merge most all recent PRs together.

PR #641, opened 2026-08-16
Support configuring when read marker updates are triggered

### LINKED ISSUE #484: Unread status not being cleared on visible buffers
When a tab/window contains multiple chats buffers (splitted) only the focused/active buffer is marked as read.

I'd expect that all visible buffers in the window are marked as read.

ulyssa:
This resolves #484 by adding a new `settings.read_receipt_trigger` option that can be set to different values to control just how in-focus the most recent message needs to be. Possible values are:

- `focused`, the default and the current behaviour: the room must be scrolled to the most recent message, and the window must be focused.
- `visible`, a slightly more relaxed condition like asked for in #484, which only requires that the room is scrolled to the most recent message and is on-screen.
- `scrollback`, an even more relaxed condition, which only requires that some portion of the scrollback is visible.
- `message`, which delays updating until the user actually sends a message. (i.e., don't consider rooms read until a response has been sent.)
- `disabled`, which disables any automatic updates to the markers, and requires using either `:unreads clear` or `:room unread clear` to clear read markers.

### DIFF
--- a/docs/iamb.5
+++ b/docs/iamb.5
@@ -229,6 +229,25 @@ Defines whether or not reactions should be shown.
 Defines whether or not reactions should be shown as their respective shortcode.
 .It Sy read_receipt_send
 Defines whether or not public read confirmations are sent.
+.It Sy read_receipt_trigger
+Defines under what conditions threads and rooms are marked read.
+Possible values are:
+.Bl -tag -width Ds
+.It Sy focused
+The room or thread must be focused and scrolled to the last message.
+This is the default value.
+.It Sy visible
+The room or thread must be on screen and scrolled to the last message.
+.It Sy scrollback
+Some portion of the scrollback for the room or thread must be on screen, but
+the room does not need to be scrolled to the last message.
+.It Sy message
+A message must be sent to the room before marking it read.
+.It Sy disabled
+Rooms and threads do not get their read markers updated, and the room
+will remain marked unread until manually cleared or another device updates
+the read marker.
+.El
 .It Sy read_receipt_display
 Defines whether or not read confirmations are displayed.
 .It Sy request_timeout
--- a/src/base.rs
+++ b/src/base.rs
@@ -93,13 +93,12 @@ use modalkit::{
     prelude::{CommandType, WordStyle},
 };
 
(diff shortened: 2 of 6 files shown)


VAWVAW:
I'm not sure there is actually a practical difference between `disabled` and `message`. I think some part of the sdk interprets a message as an implicit read marker. But I might remember this wrong and I can't find a mention of this in the spec.

ulyssa:
Ah, it does like that is the case, at least trying it out in my test room. I don't see anything about the user sending messages updating read markers in [10.6](https://spec.matrix.org/v1.19/client-server-api/#receipts) or [10.7](https://spec.matrix.org/v1.19/client-server-api/#read-and-unread-markers), so I'm guessing this is either the SDK or implicit homeserver behaviour.

ulyssa:
Okay, I believe that it is this part, the "implicit read receipts":

https://github.com/matrix-org/matrix-rust-sdk/blob/113599cba52313d13bf0094180a6a74ee6ab90bd/crates/matrix-sdk/src/event_cache/caches/read_receipts.rs#L70-L82

Which is then implemented here:

https://github.com/matrix-org/matrix-rust-sdk/blob/113599cba52313d13bf0094180a6a74ee6ab90bd/crates/matrix-sdk/src/event_cache/caches/read_receipts.rs#L235-L253

The comment says it's in the spec, but I can't actually find that precise phrase in it. I did however come across matrix-org/matrix-spec#1410, which has some relevant quotes. (There also seems to be matrix-org/matrix-spec-proposals#4005, although it's just a draft.)

I'll take a closer look later to see if it's actually possible to get `"disabled"` to work. If not, then I'll just remove it.

ulyssa:
Okay, after reading matrix-org/matrix-spec-proposals#4005 ([rendered](https://github.com/beeper/matrix-spec-proposals/blob/explicit-read-receipts-on-event-send/proposals/4005-explicit-read-receipts-on-event-send.md)), I've decided to just remove `disabled`, since it sounds like even if I made it work that it would go against the requirement that:

> When the user updates their read receipt (either by using the API or by sending an event), notifications prior to and including that event MUST be marked as read.

Thank you for pointing this out, @VAWVAW !

PR #644, opened 2026-08-18
Document how to build size-optimized binaries in the README

ulyssa:
This adds a new `release-light` cargo profile, which can be used to enable the cargo options for getting a smaller binary.

Combined with enabling the `max_level_error` feature for the `tracing` crate, I can get a 30M binary locally (as opposed to the 106M binary):

```
cargo build --profile release-light --features max_level_error
```

If I build without the `desktop` and `bundled` features, that goes down to ~28M:

```
cargo build --profile release-light --features max_level_error --no-default-features
```

Building with `codegen-units=1`, that takes it down to ~25M:

```
RUSTFLAGS="-C codegen-units=1" cargo build --profile release-light --features max_level_error --no-default-features 
```

Using `cargo-bloat`, it looks like nearly 10M comes from `libstd`, so I tried building it from source, which gets it down to ~22M:

```
RUSTFLAGS="-C codegen-units=1 -Zlocation-detail=none -Zfmt-debug=none" cargo +nightly build --profile release-light --features max_level_error --no-default-features -Z build-std=std,panic_abort -Z build-std-features="optimize_for_size" --target x86_64-unknown-linux-gnu
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -63,6 +63,12 @@ const DEFAULT_ENABLE_TITLE: bool = true;
 const DEFAULT_ENC_INDICATOR_LOC: EncryptionIndicatorLocation = EncryptionIndicatorLocation::PROMPT;
 const DEFAULT_REQ_TIMEOUT: u64 = 120;
 
+const DEFAULT_LOG_LEVEL: &str = if cfg!(feature = "max_level_error") {
+    "error"
+} else {
+    "warn"
+};
+
 const COLORS: [Color; 13] = [
     Color::Blue,
     Color::Cyan,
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -63,6 +63,12 @@ const DEFAULT_ENABLE_TITLE: bool = true;
 const DEFAULT_ENC_INDICATOR_LOC: EncryptionIndicatorLocation = EncryptionIndicatorLocation::PROMPT;
 const DEFAULT_REQ_TIMEOUT: u64 = 120;
 
+const DEFAULT_LOG_LEVEL: &str = if cfg!(feature = "max_level_error") {
+    "error"
+} else {
+    "warn"
+};
+
```
VAWVAW, line 71: I believe this is unnecessary. `max_level_error` makes sure that no `warn!` logs are even compiled in. Therefore they can never be emitted and both log levels `error` and `warn` produce the same output.

ulyssa: It's needed for when the binary is run. Without it, you get a warning message from `tracing` when starting iamb because `setup_logging` tries to set the log level to something below its minimum.

PR #651, opened 2026-08-25
Truncate usernames by grapheme instead of word boundaries

kaizoplant:
Previously, usernames truncated by user_gutter_width with word boundaries, such as display names like [somelongtexthere] or more relevant, usernames like @username:some.domain would be completely cut off after the first character. This pr makes it so that these will always just use the grapheme calculation directly.

Maybe this should be a config option? Although I do think it is usually preferable to have the username as visible as possible, it might still be useful to not always  have this behavior.

### DIFF
--- a/src/message/mod.rs
+++ b/src/message/mod.rs
@@ -68,7 +68,7 @@ use crate::{
     base::RoomInfo,
     config::ApplicationSettings,
     message::html::{StyleTree, parse_matrix_html},
-    util::{replace_emojis_in_str, space, space_span, take_width, wrapped_text},
+    util::{replace_emojis_in_str, space, space_span, take_width_grapheme, wrapped_text},
 };
 
 mod compose;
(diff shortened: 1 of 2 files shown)


VAWVAW:
I believe the `UnicodeSegmentation` functions were always relatively CPU intensive. Have you tested whether this may introduce lag on lower-end systems?

kaizoplant:
It doesn't seem to show up on the flamegraph for me so I imagine it shouldn't be too much of a problem. Also previously, it was still calling into split_word_bound_indices which I imagine was a bit worse for long usernames and display names (especially display names given it'd have to call split_word_bound_indices and then grapheme_indices after it failed), maybe it's better if it calls into something else entirely but it doesn't seem like too much of an issue rn?

VAWVAW (approved):
I see.

PR #653, opened 2026-08-25
Image Preview improvements

### LINKED ISSUE #650: Image reactions are sometimes rendered in the wrong line
If there are enough reactions to span multiple lines and image reaction is in one of the lower lines, it is rendered one line above its space.

VAWVAW:
This PR has a lot of small changes. I can break it into multiple PRs if you want.

- update `ratatui-image` to `9.0` (further updates need a modalkit release to bump `ratatui`)
  - this uses the runtime library [chafa](https://hpjansson.org/chafa/) for halfblock rendering which I have disabled by default since it doesn't improve the quality much
  - ~`settings.image_preview.protocol.type` is no longer required on windows (ratatui/ratatui-image#32)~
  - ~generating a picker from the font size has been deprecated and I have removed the setting~
- `settings.image_preview.protocol.filter` exposes the sample filter used to scale image previews
  - there is a tradeoff between image visibility and runtime
  - in my tests using [`Triangle`](https://docs.rs/image/latest/image/imageops/enum.FilterType.html#variant.Triangle) over `Nearest` greatly improves image visibility at a marginally higher runtime

fixes #650

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4471,13 +4471,14 @@ dependencies = [
 
 [[package]]
 name = "ratatui-image"
-version = "8.1.1"
+version = "9.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3ecc67e9f7d0ac69e0f712f58b1a9d5a04d8daeeb3628f4d6b67580abb88b7cb"
+checksum = "5dbebe428e366c230b2251c7091f2a56ec2bf818d96cdc807f6474428ee09040"
 dependencies = [
  "base64-simd",
  "icy_sixel",
  "image",
+ "pkg-config",
  "rand 0.8.6",
  "ratatui",
  "rustix 0.38.44",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2581,6 +2581,7 @@ dependencies = [
  "ravif",
  "rayon",
  "rgb",
+ "serde",
  "tiff",
  "zune-core",
  "zune-jpeg",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,6 @@ nom = "8.0"
 open = "5.0.0"
 rand = "0.10"
 ratatui = "0.29.0"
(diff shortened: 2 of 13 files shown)


benjajaja:
Chafa does improve a lot over halfblocks! It's just a huge pain in the ass to add, so I wouldn't bother unless you want to have "fun" with linking.

The latest version has "sliced protocols", which would be a real killer feature here, since chats scroll all the time. It allows for partial rendering, so if the image is partially out of view, you can still render the visible parts (lines).

VAWVAW:
In my test I've seen mixed results with `chafa`. Most pictures were clearer but for some it picked weird colors that made the picture harder to recognize.
I have added cargo features for anyone who wants to bother.

I have an eye on the sliced protocols but for that we need a new `modalkit` release to keep the `ratatui` version in sync.

@benjajaja What is the current state of `Picker::from_query_stdio()` on Windows? The issue I linked was closed before `9.0` was released. But the CI test times out. Is that a problem with the CI environment or should we stick to the now-deprecated `Picker::from_fontsize()` on Windows?

benjajaja:
That is a good question. I think we should stick to from_fontsize() right now, but there was recently some patch to address that, IIRC - to make from_query_stdio() use the ioctl of from_fontsize(), which seems to work on windows despite the queries not working.

I'll check it, cut a release, and report back here.

benjajaja:
To be clear, that patch change I mentioned is not in any release yet, so we should keep using from_fontsize() for the time being, for windows.

VAWVAW:
Yeah, I updated this PR to not call `from_query_stdio()` on windows. As long as `from_fontsize()` stays around until this issue is fixed, we should be fine.

VAWVAW:
I can understand not wanting to have `iamb_dbg!` but it's the same situation as with [`std::dbg!`](https://doc.rust-lang.org/std/macro.dbg.html) where it shouldn't be used in production-like code (which the `main` branch probably qualifies as).

PR #660, opened 2026-08-31
Improve initial load time by using the event cache

VAWVAW:
Insert the most recent events from the `matrix-sdk` cache into `RoomInfo` before doing network requests. This has reduced the time until the chats are ordered by `recent` by a factor of 6 for me.
Also use the event cache for backwards pagination.

A possible followup would be to replace the `RoomNeeds` with a `mpsc` channel and reduce the debounce the make going back in the scrollback a lot faster. But I ran into issues with creating futures on the run that I didn't want to deal with now.


VAWVAW:
I had some problems with this not properly loading member events. I will lock into this.

VAWVAW:
I can't reproduce my issue and this PR doesn't touch member loading, so it was probably a fluke.

(It was a problem with my `dev` branch missing an edge case.)

PR #662, opened 2026-09-01
Add support for updating user profile information

ulyssa:
This adds new `:self [name|avatar|timezone]` commands for updating user profile information. (Although as far as I can tell, I don't think any servers have actually added support for `m.tz` information yet?)

I haven't included it here, but at some point it would be good to have a `:self avatar upload` command to upload a new file from disk.

### DIFF
--- a/docs/iamb.1
+++ b/docs/iamb.1
@@ -232,7 +232,6 @@ Delete your per-room user display name.
 .It Sy ":room user name show"
 Show the display name you currently have in this room.
 .El
-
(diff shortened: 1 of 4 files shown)


VAWVAW:
Looking at #iamb-chat, 8 users have `m.tz` set.

ulyssa:
Ah, it looks like `tuwunel` added support: https://github.com/matrix-construct/tuwunel/commit/fb5a4ea96164c00a949b3e6283e0177dc92d4d2a

PR #667, opened 2026-09-02
Use `SlicedProtocol` to render images that are partially off-screen

VAWVAW:
~This is currently broken with sixel terminals.~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1020,6 +1020,7 @@ dependencies = [
  "itoa",
  "rustversion",
  "ryu",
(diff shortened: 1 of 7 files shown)


VAWVAW:
This is now mostly fixed.

The only remaining bug is, that for `sixel` the placeholder size is reported as the `settings.image_preview.size` instead of the smaller actual size of the image. We can either wait for upstream to fix this minor bug before merging or get it in the next update.

VAWVAW:
This now also works on sixel.

PR #670, opened 2026-09-03
Add MatrixRTC + LiveKit voice call support

Nightchicken:
Adds an optional `voip` feature providing voice calls over MatrixRTC, using LiveKit as the SFU.

The new `src/voip` module contains:

  - `matrix_rtc.rs`: homeserver-facing signalling. Discovers the LiveKit focus from `m.call.member` state, exchanges an OpenID token for an SFU JWT via lk-jwt-service, and publishes our own membership.
  - `livekit_session.rs`: the LiveKit room connection, the E2EE key ring, and the published microphone track.
  - `devices.rs`: audio device enumeration, matching, and remembered choices.
  - `mod.rs`: `CallManager`/`CallSession` state machine and the key exchange event types.

Calls in encrypted rooms are end-to-end encrypted. Per-participant keys are distributed over Olm-encrypted to-device messages (`io.element.call.encryption_keys`), and re-sent to late joiners. In unencrypted rooms audio is sent in the clear, matching the room.

Audio capture and playback are both handled by LiveKit's platform audio device module, so no additional audio dependency is needed.

User-facing surface is the `:call` command (`join`, `hangup`, `decline`, `mute`, `unmute`, `devices`, `device mic|speaker <name>`), a participant banner in the room view, and a room-list indicator for rooms with an active call.

The feature is off by default; without `--features voip` the build is unchanged.

Claude-Session: https://claude.ai/code/session_01MNTPSETtCAzUuLbn4fzG9a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,6 @@ on:
   push:
     branches:
       - main
-      - 'voip-**'
   pull_request:
     branches:
       - main
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,7 @@ on:
   push:
     branches:
       - main
+      - 'voip-**'
   pull_request:
     branches:
       - main
@@ -45,6 +46,67 @@ jobs:
         reporter: 'github-check'
     - name: Run tests
       run: cargo test --locked
+
+  voip:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: ubuntu-latest
+            target-dir: target
+          - platform: macos-latest
+            target-dir: target
+          # Keep the target directory short: the prebuilt libwebrtc tree
+          # unpacks deeply and blows past MAX_PATH under D:\a\iamb\iamb.
+          - platform: windows-latest
+            target-dir: C:/t
+    runs-on: ${{ matrix.platform }}
+    timeout-minutes: 90
+    env:
+      CARGO_TARGET_DIR: ${{ matrix.target-dir }}
+      SCCACHE_GHA_ENABLED: "true"
+      RUSTC_WRAPPER: "sccache"
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+      with:
+        submodules: true
+    - name: Install Rust (1.93 w/ clippy)
+      uses: dtolnay/rust-toolchain@1.93
+      with:
+          components: clippy
+    # webrtc-sys unwraps pkg_config for glib-2.0/gobject-2.0/gio-2.0, so the
+    # headers are a hard build dependency. X11/drm/gbm/VA-API are not: they are
+    # dlopened through stubs shipped in the crate.
+    - name: Install system dependencies (Linux)
+      if: runner.os == 'Linux'
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y libglib2.0-dev
+    - name: Enable long paths (Windows)
+      if: runner.os == 'Windows'
+      run: git config --system core.longpaths true
+    - name: Cache cargo registry
+      uses: actions/cache@v4
+      with:
+        path: ~/.cargo/registry
+        key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
+    # A cold build downloads and unpacks ~1.4GB of prebuilt libwebrtc into the
+    # `scratch` crate's directory under CARGO_TARGET_DIR. Cache it separately
+    # from the rest of target/, which we let cargo rebuild.
+    - name: Cache prebuilt libwebrtc
+      uses: actions/cache@v4
+      with:
+        path: ${{ matrix.target-dir }}/*/build/scratch-*/out/livekit_webrtc
+        key: ${{ runner.os }}-libwebrtc-${{ hashFiles('**/Cargo.lock') }}
+    - name: Run sccache-cache
+      uses: mozilla-actions/sccache-action@v0.0.9
+    - name: Check Clippy
+      run: cargo clippy --locked --features voip --all-targets -- --deny warnings
+    - name: Build
+      run: cargo build --locked --features voip
+    - name: Run tests
+      run: cargo test --locked --features voip
 
   nix-flake-test:
     name: Flake checks ❄️
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,12 @@ max_level_error = ["tracing/max_level_error"]
 max_level_off = ["tracing/max_level_off"]
 chafa-dyn = ["ratatui-image/chafa-dyn"]
 chafa-static = ["ratatui-image/chafa-static"]
+voip = [
+    "dep:livekit",
+    "dep:base64",
+    "dep:rustls",
+    "matrix-sdk/experimental-send-custom-to-device",
+]
 
 [build-dependencies.vergen-gitcl]
 version = "9"
@@ -31,6 +37,7 @@ features = ["build"]
 
 [dependencies]
 anyhow = "1.0"
+base64 = { version = "0.22.1", optional = true }
 bitflags = "2.3"
 chrono = "0.4"
 clap = {version = "^4.6", features = ["derive"]}
@@ -93,6 +100,25 @@ version = "0.0.26"
 
 [dependencies.matrix-sdk-base]
 version = "0.18.0"
+
+# VoIP audio + LiveKit SFU, enabled by the `voip` feature.
+[dependencies.livekit]
+version = "0.7.52"
+optional = true
+default-features = false
+# `tokio` = signal-client-tokio + tokio runtime (livekit's default backend);
+# rustls-tls-native-roots keeps us on rustls instead of native-tls/OpenSSL.
+features = ["tokio", "rustls-tls-native-roots"]
+
+# LiveKit's WebSocket TLS goes through rustls, whose automatic CryptoProvider
+# selection panics when the dependency graph enables both `ring` and `aws-lc-rs`
+# (it does, via the two reqwest majors). We install the `ring` provider as the
+# process default at startup; this dependency is only to name that provider.
+[dependencies.rustls]
+version = "0.23"
+optional = true
+default-features = false
+features = ["ring"]
 
 [dependencies.matrix-sdk]
 version = "0.18.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,11 @@ voip = [
     "matrix-sdk/experimental-send-custom-to-device",
 ]
 
+# ruma's `EventContent` derive emits `#[cfg(ruma_unstable_exhaustive_types)]`,
+# which is not a cfg we set. Declare it so `clippy -- --deny warnings` passes.
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(ruma_unstable_exhaustive_types)'] }
+
 [build-dependencies.vergen-gitcl]
 version = "9"
 default-features = false
--- a/docs/iamb.1
+++ b/docs/iamb.1
@@ -273,6 +273,28 @@ Delete your per-room user display name.
 .It Sy ":room user name show"
 Show the display name you currently have in this room.
 .El
+
+.Sh "CALL COMMANDS"
+.Bl -tag -width Ds
+.It Sy ":call"
+Join the voice call in the currently selected room, or answer a call you have been rung about.
+Your audio is encrypted when the room is encrypted, and sent in the clear when it is not.
+.It Sy ":call hangup"
+Leave the call and withdraw your membership from it.
+.It Sy ":call decline"
+Reject a call you have been rung about without joining it.
+.It Sy ":call mute"
+Mute your microphone.
+.It Sy ":call unmute"
+Unmute your microphone.
+.It Sy ":call devices"
+List the available microphones and speakers.
+.It Sy ":call device mic [device]"
+Choose the microphone by index or name.
+.It Sy ":call device speaker [device]"
+Choose the speaker by index or name.
+.El
+
 .Sh "SPACE COMMANDS"
 .Bl -tag -width Ds
 .It Sy ":space child set [room] [arguments]"
--- a/src/base.rs
+++ b/src/base.rs
@@ -109,7 +109,7 @@ use crate::{
 };
 
 #[cfg(feature = "voip")]
-use crate::voip::{CallStatus, IncomingCall, devices::DeviceKind};
+use crate::voip::{IncomingCall, devices::DeviceKind};
 
 /// The set of characters used in different Matrix IDs.
 pub const MATRIX_ID_WORD: WordStyle = WordStyle::CharSet(is_mxid_char);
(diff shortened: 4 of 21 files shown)


src/main.rs
```diff
@@ -1043,12 +1098,81 @@ async fn login_normal(
...
+        libc::dup2(saved, libc::STDERR_FILENO);
+        libc::close(saved);
+    }
+}
+
+#[cfg(not(all(unix, feature = "voip")))]
+fn silence_native_stderr() {}
+
+#[cfg(not(all(unix, feature = "voip")))]
+fn restore_native_stderr() {}
```
VAWVAW: Have you considered using other libraries that don't clutter `stderr`?

And I think this can be enabled on all builds and doesn't need to be behind a feature flag.

src/main.rs
```diff
@@ -1194,12 +1335,27 @@ fn setup_logging(settings: &ApplicationSettings) -> tracing_appender::non_blocki
         .with_env_filter(filter)
         .finish();
 
+    // `try_init` also installs the `log` compatibility bridge. LiveKit and
+    // libwebrtc log through `log` rather than `tracing` - including the sink
+    // carrying WebRTC's own C++ diagnostics - so without it everything either
+    // of them says is discarded, and a call that fails inside the media stack
+    // leaves nothing behind to debug with.
+    #[cfg(feature = "voip")]
+    subscriber.try_init().expect("setting default subscriber failed");
```
VAWVAW: Is `try_init().expect(..)` different form `init()`?

src/main.rs
```diff
@@ -1194,12 +1335,27 @@ fn setup_logging(settings: &ApplicationSettings) -> tracing_appender::non_blocki
...
+
+    #[cfg(not(feature = "voip"))]
     tracing::subscriber::set_global_default(subscriber).expect("setting default subscriber failed");
```
VAWVAW: If the init call possibly provides better logging, we should do that on all builds.

VAWVAW (commented):
I haven't looked at this in depth, but here are some of my initial thoughts:
- if you add a substantial feature flag, you should also add a CI test to make sure everything works with the feature enabled and disabled
- I'm not a fan of the conditional `use` statements. The imports are already very cluttered and I feel this makes it potentially even worse. I would use fully qualified calls whenever possible and put the imports that are necessary in their own section.

src/main.rs
Nightchicken: I fixed the issues that was causing stderr to become flooded, I will be removing that outright.

src/windows/welcome-call.md
VAWVAW, line 1: This file doesn't seem to be used.

.github/workflows/ci.yml
```diff
@@ -2,6 +2,7 @@ on:
   push:
     branches:
       - main
+      - 'voip-**'
```
VAWVAW: This needs to be removed before a merge. (And probably doesn't do anything now that the PR is opened.)

.github/workflows/ci.yml
```diff
@@ -46,6 +47,67 @@ jobs:
...
+      with:
+          components: clippy
+    # webrtc-sys unwraps pkg_config for glib-2.0/gobject-2.0/gio-2.0, so the
+    # headers are a hard build dependency. X11/drm/gbm/VA-API are not: they are
+    # dlopened through stubs shipped in the crate.
+    - name: Install system dependencies (Linux)
+      if: runner.os == 'Linux'
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y libglib2.0-dev
```
VAWVAW, line 84: This line looks like it introduces a runtime dependency. That should probably be documented in the readme.

Cargo.toml
```diff
@@ -23,6 +23,12 @@ max_level_error = ["tracing/max_level_error"]
 max_level_off = ["tracing/max_level_off"]
 chafa-dyn = ["ratatui-image/chafa-dyn"]
 chafa-static = ["ratatui-image/chafa-static"]
+voip = [
```
VAWVAW, line 26: The new `voip` feature should be mentioned in the readme.

src/main.rs
```diff
@@ -1154,11 +1208,24 @@ async fn run(settings: ApplicationSettings) -> IambResult<()> {
...
+        worker.call_hangup_on_exit(call.room_id);
+    }
 
     // Clean up the terminal on exit.
     restore_tty(enable_enhanced_keys, enable_mouse);
 
+    res?;
+
     Ok(())
 }
```
VAWVAW:
```suggestion
    res.map_err(UIError::from)
}
```

.github/workflows/ci.yml
VAWVAW: And maybe add `pkgs-config` and `glib` to the `flake.nix` devShell packages (line 130).

docs/iamb.1
```diff
@@ -273,6 +273,28 @@ Delete your per-room user display name.
 .It Sy ":room user name show"
 Show the display name you currently have in this room.
 .El
+
+.Sh "CALL COMMANDS"
+.Bl -tag -width Ds
+.It Sy ":call"
+Join the voice call in the currently selected room, or answer a call you have been rung about.
+Your audio is encrypted when the room is encrypted, and sent in the clear when it is not.
```
VAWVAW, line 281: I feel like it might make sense to have this option named something like `:call join`. That would leave the possibility to have a call list on `:call` like with `:verify`.

src/windows/mod.rs
```diff
@@ -85,6 +85,9 @@ use self::welcome::WelcomeState;
 use crate::message::MessageTimeStamp;
 use feruca::Collator;
 
+#[cfg(feature = "voip")]
+use crate::base::CallAction;
+
```
VAWVAW: What I meant with the imports was that you could completely remove this import and just use `crate::base::CallAction` instead of `CallAction` everywhere in this module.

That would reduce the number of `cfg`s. But that is my personal preference.

src/base.rs
```diff
@@ -1930,13 +2016,24 @@ pub struct ChatStore {
 
     /// Notifications that should be dismissed when the user opens the room.
     pub open_notifications: HashMap<OwnedRoomId, Vec<NotificationHandle>>,
+
+    /// The call we are currently joined to, if any.
+    ///
+    /// Read-only here: the worker owns this state and is its only writer.
+    #[cfg(feature = "voip")]
+    pub call_status: CallStatus,
```
VAWVAW: I don't see why this needs to have a clone in `ChatStore` instead of accessing it through `store.application.worker.call_status`.

src/voip/mod.rs
```diff
@@ -0,0 +1,1206 @@
+//! # Voice/Video calls (VoIP)
+//!
+//! This module implements voice calls for iamb using [LiveKit] for the media
```
VAWVAW, line 3: `cargo doc` complains about this being a dead link. (And a few others in this file.)

VAWVAW (commented):
I still haven't had time to look fully into this but here are a few more thoughts.

Does this PR only support [MSC4075](https://github.com/matrix-org/matrix-spec-proposals/pull/4075) or also the [stable voip](https://spec.matrix.org/v1.19/client-server-api/#voice-over-ip) spec?

src/worker.rs
```diff
@@ -1568,4 +2356,392 @@ impl ClientWorker {
...
+        }
+
+        PlatformAudio::new().map_err(call_error)
+    }
+
+    /// Show the audio devices available for calls.
+    #[cfg(feature = "voip")]
+    fn call_devices(&self) -> IambResult<EditInfo> {
+        let audio = self.audio_handle()?;
+        let prefs = self.settings.read_voip_devices();
```
VAWVAW: I'm not sure reading the file every time is better than caching the preferences somewhere in `ApplicationSettings` and writing them once on change.

src/voip/devices.rs
```diff
@@ -0,0 +1,357 @@
...
+        let chosen = prefs.get(kind);
+
+        for device in devices {
+            let mark = if Some(device.name.as_str()) == chosen {
+                "  (selected)"
+            } else {
+                ""
+            };
+
+            out.push_str(&format!("  {}  {}{mark}\n", device.index, device.name));
```
VAWVAW, line 138: nitpick (feel free to ignore): Using `write!(..)` instead of `format!(..)` removes one unnecessary allocation.

docs/iamb.1
Nightchicken: There is a :call join, but I had ":call" call the same join command since ":call" felt more idiomatic to imitate a call.

src/base.rs
```diff
@@ -628,6 +631,30 @@ pub enum KeysAction {
...
+    /// Leave the call in the currently focused room.
+    Hangup,
+    /// Decline a call we have been rung about, without joining it.
+    Decline,
+    /// Set the local microphone mute state (`true` = muted).
+    Mute(bool),
+    /// Show the available audio devices.
+    Devices,
+    /// Choose an audio device, by index or by name.
+    SetDevice(DeviceKind, String),
```
VAWVAW, line 655: I think there should be an option for an interactive device select using `keybindings::dialog::MultiChoice`.

src/worker.rs
```diff
@@ -717,6 +787,96 @@ impl<T> ClientReply<T> {
...
+///
+/// `latest_encryption_state` fetches `m.room.encryption` when the state never
+/// reached us, which matters because `:call` works from the room list, on rooms
+/// we may never have opened. If even that fails we assume encryption: guessing
+/// wrong that way costs audio, while guessing wrong the other way puts our
+/// media on the wire in the clear.
+#[cfg(feature = "voip")]
+async fn call_media_encrypted(room: &MatrixRoom) -> bool {
+    match room.latest_encryption_state().await {
+        Ok(state) => !matches!(state, EncryptionState::NotEncrypted),
```
VAWVAW: I think adding an extra match arm here would be more readable than the `matches!()`.

Nightchicken:
This only implements the newer MatrixRTC, it does not have code for the legacy VOIP 1:1. The clients I have checked all have the new API. Including element desktop and element X.

src/worker.rs
```diff
@@ -1568,4 +2356,392 @@ impl ClientWorker {
...
+        }
+
+        let Some(room) = self.client.get_room(&room_id) else {
+            return Err(IambError::UnknownRoom(room_id).into());
+        };
+
+        let encrypted = call_media_encrypted(&room).await;
+
+        if !encrypted {
+            tracing::warn!(%room_id, "the room is unencrypted, so our media is too");
```
VAWVAW: I think a `warn` might be too verbose here. An unencrypted call in an unencrypted room is probably to be expected.

VAWVAW:
I am a bit confused by the different matrix specs for doing voice. There seem to be:
- 1:1 voip calls ([stable](https://spec.matrix.org/v1.19/client-server-api/#voice-over-ip))
- voip group calls ([MSC3401](https://github.com/matrix-org/matrix-spec-proposals/pull/3401))
- MatrixRTC
  - this seems to supersede MSC3401 and is made up from multiple MSCs
  - core: [MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143)
  - calls: [MSC4195](https://github.com/matrix-org/matrix-spec-proposals/pull/4195)
  - livekit: [MSC4196](https://github.com/matrix-org/matrix-spec-proposals/pull/4196)
  - call notifications: [MSC4075](https://github.com/matrix-org/matrix-spec-proposals/pull/4075)

I couldn't find clear information on this but element seems to support a combination of the latter two. Could you please state and link the documentation you used to implement this PR?

Nightchicken:
I was not familiar with rusts docs with their links, I was using markdown links previously, my next commit will contain working links to the related resources. 

until that here is the related docs

[MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143) 
[MSC4195](https://github.com/matrix-org/matrix-spec-proposals/pull/4195) 
[MSC4196](https://github.com/matrix-org/matrix-spec-proposals/pull/4196)
[MSC3401](https://github.com/matrix-org/matrix-spec-proposals/pull/3401) 
[MSC4075](https://github.com/matrix-org/matrix-spec-proposals/pull/4075)
[MSC3757](https://github.com/matrix-org/matrix-spec-proposals/pull/3757)
[MSC4310](https://github.com/matrix-org/matrix-spec-proposals/pull/4310) 

Some of these are closed and rejected, while others are obsolete, but the use of them did make some of the code easier to implement. For example the MSC3757 was rejected, but implementing it solved some issues I had with key verification. 

It was difficult to track down the exact new spec. At least what was required to make it work properly with element's client. These were the set that made things work.

There are some specs that would be great to add, like delayed events in case of a crash, but that would have to interact with the code base more.

VAWVAW:
Ok, the whole ecosystem of call on matrix is a mess.

I will refer to the combination of [MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143), [MSC4195](https://github.com/matrix-org/matrix-spec-proposals/pull/4195), and [MSC4196](https://github.com/matrix-org/matrix-spec-proposals/pull/4196) as MatrixRTC.

The spec discussions seem to favor MatrixRTC over [MSC3401](https://github.com/matrix-org/matrix-spec-proposals/pull/3401). ([MSC4075](https://github.com/matrix-org/matrix-spec-proposals/pull/4075) changed its focus to MatrixRTC.)

MatrixRTC is still very actively discussed and the only up-to-date implementations seem to be [Element Call](https://github.com/element-hq/element-call/) and synapse with [lk-jwt-service](https://github.com/element-hq/lk-jwt-service). There is work being done to add support for base features of MatrixRTC to the matrix-sdk ([1], [2], [3]).

Element Web seems to support a version of MSC3401 but isn't listed as an implementer. It also uses an old version of both MSC4075 and the now-closed MSC4143. The same goes for Element-X (android).

Here are the options I see for going forward:
1. Go for maximum compatibility
    - Implement the combination of old/closed MSCs that Element Web does (this PR)
2. Implement the currently newest version
    - Would only be compatible with Element Call for now (as far as I can see)
    - Element Web will probably move over at some point
3. Wait until the discussion has settled a bit
    - All the MatrixRTC MSCs (except MSC4075) are currently in their final review period meaning that there soon is a fixed spec to develop against

@Nightchicken, @ulyssa Do you see anything I missed and what are your opinions on this matter?

I don't like option 1 since we would add new support for (almost) obsolete standards and about-to-be obsolete features.

[1]: https://github.com/BillCarsonFr/matrix-rust-rtc/issues/31
[2]: https://github.com/matrix-org/matrix-rust-sdk/pull/7014
[3]: https://github.com/matrix-org/matrix-rust-sdk/issues/3295

Nightchicken:
I avoided adding the voip 1:1 because the standard is pretty much obsolete. I would prefer that we do not work on something that will be discontinued soon. I have tried to implement the new MatrixRTC standard to the best of my ability.

I am okay with going with option 3, but as of right now this PR is more or less option 2. 
There are people in the community that have been using this implementation and it has been working for them. So it is at least functional. 

As far as Element X (android). It has both the MatrixRTC and voip implementations. Which would still make this compatible.

VAWVAW:
I don't agree that this PR falls under option 2. I think it is clearly option 1 as it implements MSC3401 and the removed MSC4310.

I believe the best way forward is to wait until matrix-org/matrix-rust-sdk#7014 is merged and then implement MatrixRTC based on that and possibly ruma/ruma#2555.

Nightchicken:
Honestly that seems fair enough.

PR #683, opened 2026-09-11
Allow user's to specify a default register for editing

### LINKED ISSUE #410: [Feature Request] User-Specified Registers
Is it possible to tell iamb to use the unnamed register when yanking/deleting? 
It would be cool if it were possible to specify this in the config.toml

something like:
```toml
[settings.clipboard]
default_register = "unnamedplus"
```

If one wanted, they might specify iamb to always use the `a` register or the `q` one for whatever reason.

ulyssa:
This updates to the latest modalkit release and adds a new `default_register` configuration option to take advantage of ulyssa/modalkit#217. I have restricted this to `a-z`, `_`, `+`, `*`, and `"` for now, so that folks don't get confused by trying to use a read-only register, or get confused by the automatically updated registers like `0-9`, which would stomp over whatever they copy or delete.

resolves #410


fuguesoft:
poggers

Issue #571, opened 2026-02-11
Terminal state in not properly restored on exit


piko-piko:
I'm using kitty as my terminal emulator. When i close iamb and then open nano, the Fx keys and ctrl+<key> shortcuts are not being recognized properly (for example F2 instead of saving tries to include a file).

Vaishnav-Sabari-Girish:
This must be due to an ANSI thing. I face that a lot too when I exit TUI programs like `durdraw`, your keybinds are being read as the ANSI codes instead of the actual keybinds. 

This is solved by using the `reset` command.

PR #688, opened 2026-09-12
Disable terminal alt screen before opening the external editor

### LINKED ISSUE #577: `:editor` destroys terminal scrollback
Normally iamb operates on an alternate terminal screen meaning that the terminal scrollback history is restored after exit (like with `less(1)`).

Using `:editor` the editor is opened in the terminal (`nvim` in my case). This seems to mess with the alternate screen leading to the scrollback being overwritten.

I assume this is because `nvim` enters and then exits the alternate screen. Since it is already entered the enter does silently not work but the exit does work. Then iamb assumes it is still on the alternate screen and overwrites the scrollback.

To solve this we need to leave the alternate screen (and probably call `restore_tty()`) before calling https://github.com/ulyssa/iamb/blob/93fc47d019cd6a9d56f163aa6ba819ef1fd309d6/src/windows/room/chat.rs#L551-L554 and reenter it (or `setup_tty()`) afterwards.

Nightchicken:
Refactor to TTY commands to add them into utils. Created a command for suspending tty for the editor. Leaves alt mode and re-enters upon editor leaving.

fixes #577


src/main.rs
```diff
@@ -1089,31 +1021,24 @@ async fn run(settings: ApplicationSettings) -> IambResult<()> {
...
+    let hook_settings = settings.clone();
     std::panic::set_hook(Box::new(move |panic_info| {
-        restore_tty(enable_enhanced_keys, enable_mouse);
+        restore_tty(&hook_settings);
         orig_hook(panic_info);
         process::exit(1);
     }));
 
     // And finally, start running the terminal UI.
+    let tty_settings = settings.clone();
```
VAWVAW: I would prefer not having to clone the settings just to access two bools. It seems like overkill and the clone is not completely cheap.

Nightchicken: The clones were "cleaner" looking, but does have serious performance issues compared to just passing the bools, changed.

VAWVAW:
This fixes the overwriting issue.

But testing this with `EDITOR=nano cargo run` the nano that is opened by `:editor` has the same problems as reported in #571 (in both kitty and foot). Could you maybe look into that?

PR #691, opened 2026-09-13
Support implicit read receipts

### LINKED ISSUE #690: Support implicit read receipts
If a user sends a message, that and all previous messages should be treated as read. (The same as if the user had sent a read receipt for their own message.)

Also a user might have two read receipts (main thread and unthreaded) depending on which combination of clients they use. It might be possible to deduplicate that in the same fix.

TayfurYldz:
## Summary

This fixes #690 by separating the current user's outgoing read progress from the receipts displayed in the timeline, and by deriving implicit displayed receipts from timeline events.

The initial implementation treated a user's own messages as outgoing read progress. Based on review feedback, implicit receipts now update the displayed user receipt state for the sender of each event instead.

## Behavior

- implicit receipts are recorded for every event sender, not only the current user
- message, sticker, encrypted, and state events can advance the sender's displayed receipt
- backpagination cannot move an implicit receipt backwards
- thread events preserve their `ReceiptThread::Thread(root)` context
- received/display receipts remain separate from the current user's outgoing active receipt state
- explicit `fully_read` progress continues to update the outgoing receipt state used for server synchronization
- `Main` outgoing progress does not regress behind an existing `Unthreaded` receipt

## Testing

Added regression coverage for:

- implicit displayed receipts for another message sender
- preventing implicit receipts from moving backwards during backpagination
- preserving thread receipt context
- state-event implicit receipts
- keeping received/display receipts separate from outgoing active receipts
- preventing outgoing active receipts from moving backwards
- preserving `Main` / `Unthreaded` ordering

The tests reuse the existing `mock_room1_message` fixture where applicable.

Validated with:

- `cargo +nightly fmt --all`
- the project's import-format check
- `git diff --check`
- `cargo test --all-targets`

Fixes #690


src/base.rs
VAWVAW, line 1577: The implicit events should also be set on other message types: sticker, encryption, state events. (I'm not sure about state events but there seem to be receipts on these as well.)

src/base.rs
```diff
@@ -1553,6 +1556,9 @@ impl RoomInfo {
         settings: &ApplicationSettings,
         previews: &mut PreviewManager,
     ) {
+        let implicit_receipt =
+            (ev.sender() == &*settings.profile.user_id).then(|| ev.event_id().to_owned());
+
```
VAWVAW: The implicit receipts should work for every sender and not just the current room member.

src/base.rs
```diff
@@ -1603,6 +1613,80 @@ impl RoomInfo {
...
+    fn set_active_receipt(
+        &mut self,
+        thread: ReceiptThread,
+        user_id: &UserId,
+        event_id: OwnedEventId,
+    ) {
+        if self.active_receipt_can_advance(&thread, user_id, &event_id) {
+            self.active_receipts.insert(thread, event_id);
+        }
+    }
```
VAWVAW: My opinion is that functions should probably be inlined if the body is shorter than the header.

src/base.rs
```diff
@@ -1603,6 +1613,80 @@ impl RoomInfo {
...
+
+    fn set_implicit_receipt(&mut self, user_id: &UserId, event_id: OwnedEventId) {
+        let thread = match self.keys.get(&event_id) {
+            Some(EventLocation::Message(None, _)) => ReceiptThread::Main,
+            Some(EventLocation::Message(Some(root), _)) => ReceiptThread::Thread(root.clone()),
+            _ => return,
+        };
+
+        self.set_active_receipt(thread, user_id, event_id);
```
VAWVAW: The implicit receipt does not apply to the active receipt that is shared with the server but to the user receipts that are displayed in the timeline.

src/base.rs
```diff
@@ -2285,6 +2367,25 @@ pub mod tests {
...
+            ("sender".to_owned(), serde_json::to_value(sender).unwrap()),
+            (
+                "origin_server_ts".to_owned(),
+                serde_json::to_value(MilliSecondsSinceUnixEpoch::now()).unwrap(),
+            ),
+            ("room_id".to_owned(), Value::String("!foo:example.org".into())),
+        ])))
+        .unwrap()
+    }
+
```
VAWVAW, line 2366: You should probably use `mock_room1_message` from `src/tests.rs` for that or factor the common stuff into a new function.

VAWVAW (changes requested):

src/base.rs
TayfurYldz: Addressed. Implicit display receipts are now updated for room messages, stickers, encrypted events, and state events using each event's sender. I also added regression coverage for the state-event path.

src/base.rs
TayfurYldz: Fixed. The implicit receipt path now uses the sender from each event, so it applies to every sender rather than only the current user. The regression test covers another room member advancing their displayed receipt.

src/base.rs
TayfurYldz: Agreed. I removed the short wrapper and inlined that logic at the call site to keep the receipt path simpler.

src/base.rs
TayfurYldz: Fixed. Implicit receipts now update the displayed `user_receipts` / `event_receipts` state, while `active_receipts` is kept separate and only represents outgoing read progress synchronized to the server. I also added regression coverage to make sure received/display receipts do not become outgoing receipts.

src/base.rs
TayfurYldz: Done. The test helper now delegates to the existing `mock_room1_message` fixture from `src/tests.rs` instead of duplicating the event construction logic.

src/base.rs
```diff
@@ -1646,11 +1748,9 @@ impl RoomInfo {
...
     pub fn receipts<'a>(
         &'a self,
-        user_id: &'a UserId,
+        _user_id: &'a UserId,
     ) -> impl Iterator<Item = (&'a ReceiptThread, &'a OwnedEventId)> + 'a {
-        self.user_receipts
-            .iter()
-            .filter_map(move |(t, rs)| rs.get(user_id).map(|r| (t, r)))
+        self.active_receipts.iter()
     }
```
VAWVAW: If this function ignores its parameter, the signature should reflect that. It could be named `own_receipts()`.

src/base.rs
```diff
@@ -1630,7 +1729,10 @@ impl RoomInfo {
             .next_back();
 
         if let Some(event_id) = event_id {
-            self.set_receipt(thread, user_id, event_id.to_owned());
+            let event_id = event_id.to_owned();
+            if self.active_receipt_can_advance(&thread, &user_id, &event_id) {
```
VAWVAW: I do not understand why this check is necessary. `fully_read` should always produce the newest event id in the thread. That should not be touched by the implicit receipts, right?

Please explain your reasoning behind this check.

VAWVAW (changes requested):
The usage of `info.set_receipt` in `windows/room/mod.rs` needs to be replaced to change the active receipt.

(Also to what manner did you use AI in the creation of this PR?)

TayfurYldz:
Addressed the latest review in 66b7c6:

- renamed the outgoing iterator to `own_receipts()` so its signature
matches its semantics
- removed the extra active-receipt advance check; `fully_read` now
uses the newest event it already selects
- changed `SetUnread(false)` to update the outgoing active receipt
through `set_active_receipt`

I validated the revision with `cargo +nightly fmt --all`, the
repository import-format check, `git diff --check`, and `cargo test
--all-targets`.

For the AI question: I used AI as a pair-programming aid for codebase
navigation, reasoning through the displayed-vs-outgoing receipt state
separation, and drafting/reviewing the patch and regression tests. I
reviewed the resulting changes against the existing code paths and
validated them with the checks above.

TayfurYldz:
On the AI question: I used an AI coding assistant as an implementation
aid to help inspect the receipt flow, reason about the relevant call
sites, and draft parts of the changes/tests. I reviewed and adjusted
the resulting code against the project’s existing receipt model and
incorporated the review feedback manually. The latest follow-up is in
`66b7c68`, including the active-receipt call-site correction and
removal of the unnecessary advance check.

src/base.rs
```diff
@@ -1576,6 +1576,7 @@ impl RoomInfo {
         }
 
         self.insert(ev);
+        self.set_implicit_receipt(sender, event_id);
```
VAWVAW: Adding this to `insert_with_preview` also triggers on edits. `insert_message` and `insert_thread` would be a slightly better fit but it doesn't matter that much because the event location gets filtered in `set_implicit_receipt`.

src/base.rs
```diff
@@ -1616,6 +1617,42 @@ impl RoomInfo {
...
+        if let Some(old_event_id) =
+            self.user_receipts.get(&thread).and_then(|receipts| receipts.get(&user_id))
+        {
+            let Some(old_key) = self.receipt_key(old_event_id) else {
+                return;
+            };
+            if event_key <= *old_key {
+                return;
+            }
+        }
```
VAWVAW: At least some of this check needs to be moved to `set_receipt`.

These cases should be covered (and probably tested):
- event receipts are received after messages (don't override implicit receipt with older explicit receipt)
- messages are received after event receipts (explicit receipt has no message to attach to but the implicit receipt that will be computed is older than the explicit receipt; in this case the thread cannot be inferred)

VAWVAW (changes requested):

src/base.rs
TayfurYldz: Moved the implicit receipt update out of `insert_with_preview` and into `insert_message` / `insert_thread`, after the event location has been recorded. That keeps edits out of the path and preserves the thread context at the insertion point.

src/base.rs
TayfurYldz: Addressed in `0d9f685`. `set_receipt` now refuses to move a loaded receipt backwards when both event positions are known, and the implicit path preserves an explicit receipt whose event has not loaded yet because its ordering cannot be proven. Added regressions for both arrival orders. `cargo +nightly fmt --all -- --check`, `git diff --check`, and `cargo test --locked` all pass (102 tests).

src/base.rs
```diff
@@ -2357,6 +2423,155 @@ pub mod tests {
...
+        );
+        let newer = mock_room_message_event(
+            RoomMessageEventContent::text_plain("newer"),
+            TEST_USER2.clone(),
+            MSG5_KEY.clone(),
+        );
+
+        info.insert_message(older);
+        info.insert_message(newer);
+        info.set_receipt(ReceiptThread::Main, TEST_USER2.clone(), MSG2_EVID.clone());
```
VAWVAW, line 2660: One case I hit while testing was like this but in the order:
```rust
info.set_receipt(ReceiptThread::Main, TEST_USER2.clone(), MSG2_EVID.clone());
info.insert_message(newer);
info.insert_message(older);
```
This resulted in the receipt being on the older message.

VAWVAW (changes requested):
(note: force-pushing makes reviewing your changes a bit harder.)

src/base.rs
TayfurYldz: Reproduced that exact arrival order before changing production code: explicit receipt on MSG2, then newer MSG5, then older MSG2 left the displayed receipt on MSG2. Fixed in a5fff75 without rewriting history: when a previously-unloaded explicit receipt event arrives, the affected user receipts are reconciled against that user’s newest loaded event in the same receipt thread, then the existing ordering guard decides whether to advance. Added the exact regression. Local validation: cargo clippy --all-targets --locked -- -D warnings; cargo test --locked (103/103); nightly rustfmt check; import-format check; git diff --check — all pass.

VAWVAW (approved):

TayfurYldz:
Resolved the current `main` conflict without force-pushing. Upstream #704 replaced the polled outgoing-receipt state with the new push-based `Requester::send_receipt` path, so the merge keeps that upstream implementation and retains this PR only for implicit/displayed receipt derivation and monotonic receipt ordering.

I also adjusted the new trackbar regression to use a fresh room for its second independent marker case, since receipt updates are now monotonic.

Validation on the merged tree: `cargo +nightly fmt --all`, `git diff --check`, and `cargo test --all-targets` (113 passed). The merge commit is `c5d2a15` and is Verified.

PR #692, opened 2026-09-14
Support pinned events

### LINKED ISSUE #626: [Feat] Pinned events
I couldn't find any mentions of pinned events (`m.room.pinned_events`) in the docs or in any other issue on this repo. Would adding support for them be a good idea?

Nightchicken:
Added commands for pinning and unpinning events
`:pin` and `:unpin`
Window screen for pinned messages under `:pinned` which contains events within room

pressing enter on an even within the window jumps cursor to that event. 
There are currently limits for how far back in history it can grab, events that are more than ~1500 messages will fail to load from scrollback, or if a time limit of 2 minutes occurs.

#626


Nightchicken:
Moving stuff to fit with suggestion of having pinned messages be stored in scrollback with hidden tags.

ulyssa:
I think this is in a good state for merging as a starting point, but ideally we will want to figure out a way to get the messages to render using the `show`/`show_with_preview` methods. That's going to take some work in `modalkit-ratatui` first though, so that the `ListItem` can render images, and so that the lifetimes it uses on `ListItem::show` allow the return value to borrow from the store.

PR #695, opened 2026-09-14
Support opening location messages with :open

### LINKED ISSUE #263: Support opening location messages
[MSC3488](https://github.com/matrix-org/matrix-spec-proposals/pull/3488) allows sending `m.location` messages which contain a `geo:` URL that the `:open` command should open. (`:open` could try converting them to an OpenStreetMap URL and opening it that way, but at least on my laptop, it looks like `xdg-open` will already take care of converting `geo:` URLs to Google Maps and opening them in a browser.)

TianHengZhuang:
`:open` on a location message previously failed with a `NoAttachment` error, since location messages carry a `geo:` URI rather than an attachment.

This change recognizes location messages and, when `:open` is used, hands their `geo:` URI to the system handler (`open_command` tunable, or the platform default), so that the desktop can open it with the map application of the user's choosing. `:download` keeps reporting `NoAttachment`, as there is nothing to save.

The manual page entry for `:open` was updated, and a unit test covers the new helper.

Closes #263


docs/iamb.1
```diff
@@ -106,6 +106,9 @@ Request a new verification with the specified user.
 Download an attachment from the selected message and save it to the optional path.
 .It Sy ":open [path]"
 Download and then open an attachment, or open a link in a message.
+Location messages are instead handed to the system handler for their
+.Dq geo:
+URI, since they have no attachment that can be downloaded.
```
VAWVAW: I'm not sure that this need a documentation change. In my opinion that just extends the "open a link" to work with more messages.

VAWVAW (approved):

docs/iamb.1
ulyssa: +1. I've moved the `geo:` comment to the docs for `open_command`, where folks overriding the config with a custom script can see it.

PR #696, opened 2026-09-14
Add `:scratch` for opening a scratch buffer

ulyssa:
Copying text in tmux or screen from iamb to vim, or just using the current message bar is usually sufficient for most temporary edits, but it would be helpful to have a space inside iamb for quick temporary edits. I've added `:scratch` to support this.

This should also help folks with confirming any editing bugs.


VAWVAW (commented):
You missed setting the `ignorecase` flag on the textbox state.

VAWVAW:
Should this copy the content of the active msgbar like `:editor`?

PR #706, opened 2026-09-16
Add `:verify recover` to verify with a recovery key

### LINKED ISSUE #705: Feature request: Allow device verification with backup key
Courtesy of Matrix user @void:frei.chat.

In order to verify an iamb session, it is required to do so with another client or device. Support for device verification with a backup key allows users to start using iamb without requiring another client.

maskjelly:
Closes #705.

A session can only be verified from another client or device today. The SDK's `recovery().recover()` imports the account's secrets from secret storage, and importing the self-signing key is what signs this device and marks it verified — so a recovery key is enough to verify a fresh session, with no second client involved.

**The command:** `:verify recover <recovery key>`.

Recovery keys are displayed in space-separated groups and the command parser splits on whitespace, so the arguments after `recover` are rejoined with single spaces before the key is handed over; a single-token key works the same way. `:verify recover` with no key is the same argument error the other subcommands produce.

**The result is read back, not assumed.** Recovering imports whatever secrets the account holds, and not every account stores a self-signing key in secret storage, so after `recover()` the handler asks the SDK for this device and reports what actually happened:

- the device comes back cross-signed → *"Recovered encryption secrets; this session is now verified"*;
- it does not → *"Recovered encryption secrets, but this session is not verified (no self-signing key in secret storage)"*;
- the device cannot be read back → the plain *"Recovered encryption secrets"*.

The key is never logged, formatted, or stored — it is passed by reference straight to `recover()` (itself `#[instrument(skip_all)]` in the SDK) and dropped.

**Changes:** `src/commands.rs` (parsing + tests), `src/base.rs` (`IambAction::Recover`, the four `ApplicationAction` match arms, an `IambError::RecoveryError` conversion mirroring `VerificationRequestError`), `src/main.rs` (dispatch), `src/verifications.rs` (`iamb_recover`), `src/completions.rs` (`recover` subcommand; a recovery key is not a flow ID so it does not complete those), `docs/iamb.1` and `src/windows/welcome.md`.

**Tests:** parsing tests beside `test_cmd_verify` cover the missing key, a single-token key, and a spaced key being rejoined (`:verify recover AAAA BBBB CCCC` carries `AAAA BBBB CCCC`). The handler needs a live account, so it stays thin — the manual check is: on a fresh session with secret storage set up, run `:verify recover <key>` and confirm the session shows as verified elsewhere; a bad key surfaces through the normal error path.

**Gates:** `cargo +nightly fmt --all -- --check`, `check-imports-format.sh` over every touched file, `cargo clippy --workspace --all-targets --locked` (zero warnings), `cargo test --locked` (95 passed) — all green.

**Not done:** no passphrase-specific UX (the SDK accepts a passphrase too, but the issue asks for the backup key); nothing was added to refresh UI state after recovery, because iamb does not display verification state anywhere for the existing `:verify` flows either — if that ever changes it should update through the SDK's own state streams.


src/completions.rs
```diff
@@ -371,11 +371,14 @@ fn complete_iamb_verify(args: Vec<String>, store: &ChatStore) -> Vec<String> {
         "cancel",
         "missmatch",
         "emoji",
+        "recover",
     ];
     match args.len() {
         1 => complete_choices(&args[0], &subcmds),
         2 if args[0] == "request" => complete_users(&args[1], store),
-        2 if subcmds.contains(&args[0].as_str()) => complete_verification(&args[1], store),
+        2 if subcmds.contains(&args[0].as_str()) && args[0] != "recover" => {
```
VAWVAW: (nitpick:) I would find it more readable if the second check was a line above that just returns `vec![]`.

PR #708, opened 2026-09-16
Unify the the room list representation

VAWVAW:
The api of `GenericChatItem`, `RoomItem`, `DirectItem` and `SpaceItem` were almost identical and are now unified under a single `GenericRoomItem`.

This will make future improvements like showing unjoined rooms in spaces easier.


ulyssa:
Nice, this is much simpler! :100:

